PR #12, opened 2017-02-19
battery@schorschii: Version 2.2

schorschii:
- added the possibility to resize the battery desklet


ChooChooAl:
Will not install.   Neither copy from website or from Synaptic (both 2.2).   Get following errors in LookinGlass:
error t=2017-03-20T17:01:04Z Initial settings schema file is not valid for battery@schorschii
error t=2017-03-20T17:01:04Z Problem initializing settings for battery@schorschii
error t=2017-03-20T17:01:04Z this.settings.bind is not a function
error t=2017-03-20T17:01:04Z [Desklet "battery@schorschii"]: Failed to evaluate 'main' function on desklet: battery@schorschii/29
error t=2017-03-20T17:01:04Z Could not load desklet battery@schorschii

collinss:
@ChooChooAl Are you using a version of Cinnamon < 3.2? That would certainly explain the errors you are getting.

ChooChooAl:
Yes.  I'm still on 2.8 it seems, mint 17.3.    Odd that it's in the repositories.    What version can I downgrade to that will work?

schorschii:
Unfortunately, for now there is no version that will work on versions below LM 18. I started developing this desklet not long ago, only tested it on Mint 18.
Currently I'm not sure what the exact problem is, but I will figure that out.

Odyseus:
Hello, everybody.

@schorschii: You most certainly are not using Mint 18. Otherwise, you will be facing the exact same problem that @ChooChooAl is facing.

The desklet has two problems:

- You are using **this.settings.bind**, which is only available on Cinnamon 3.2.x (LM 18.1). Switch to use **this.settings.bindProperty** and add a binding direction as a fist argument and you are done.

Example:

```javascript
this.settings.bindProperty(Settings.BindingDirection.IN, "devfile_capacity", "devfile_capacity", this.on_setting_changed);
```

- The second problem is in the **settings-schema.json** file. The use of *switch* elements was introduced with Cinnamon 3.2.x. Switch to the use of *checkbox* element.

These two changes will make the desklet 100% compatible with Cinnamon 2.8, 3.0.7 and 3.2. Well, at least it will make the desklet able to be placed on the desktop. I have no idea if it will be functional because I don't have a device with a battery to test it on.

ChooChooAl:
Thanks.   It does seem to work.   Not sure of all the features it has but unplugging shows % and plugging back in shows the little plug (default settings).     Checking the custom label name does not display but that's a minor issue and I don't really care since I only have one battery anyway.

Thanks again.   For me, this is a closed issue.

schorschii:
@Odyseus Thanks for your solution. Of course you are right, I'm using Mint 18.1. I'll improve my desklet and commit it soon.
Also thanks @ChooChooAl for reporting.

PR #18, opened 2017-03-22
battery@schorschii: Version 2.3

schorschii:
- works now with cinnamon versions below 3.2.x


collinss:
@schorschii Are you familiar with multi-version support for xlets? This would allow you to keep the newer stuff, but still support older versions of Cinnamon. You can find more info at http://developer.linuxmint.com/reference/git/cinnamon-tutorials/xlet-versioning.html.

schorschii:
@collinss Thanks, I didn't know that this is possible. But I think in that case it's not really necessary, or did I missed something? I mean with these adjustments there is no function loss for 3.2.8 users.

PR #19, opened 2017-04-02
Added new desklet clock@schorschii

schorschii:
Added my analog clock desklet


collinss:
The "dangerous" property should have been removed from metadata.json before this was merged. It serves a valuable purpose, which currently is undermined by overriding it in the metadata.json file. Though there is nothing in this desklet that seems to be a problem right now, any dangerous function calls added in the future would not be flagged as they should be. Also the "last-edited" property, though harmless to include, is not needed here as it is automatically generated by the server and so will be overwritten.

brownsr:
@schorschii Would be grateful if you could review the 'dangerous' property, as mentioned above, and submit a PR to change it if you don't think it is needed.  Also I am seeing a warning message in the logs:

Cjs-Message: JS WARNING: [/home/simon/.local/share/cinnamon/desklets/clock@schorschii/desklet.js 106]: assignment to undeclared variable seconds_deg

collinss:
https://github.com/linuxmint/cinnamon-spices-desklets/commit/243524d1ff0a48ff91f615d9351e2c882b96785a

schorschii:
@collinss , @brownsr Thanks for this improvement. This was not intentionally. I've just copied the metadata.json from my previous battery desklet. I'll take care of this in future.

I will correct the JS warning with the other issues created from @NikoKrause in the next commit.

PR #24, opened 2017-04-07
[analog-clock@cobinja.de] Switch to multiversion support for future development

Cobinja:
... and remove "dangerous: false"-flag (I didn't change my opinion that it's not dangerous, but whatever)


collinss:
@Cobinja the dangerous property is added by Cinnamon-settings applets the first time it opens after installing an applet to flag any potentially problematic function calls. The reason this was added is that there were a number of 3rd party applets that were causing freezes and crashes. The user would then report that as a bug to Cinnamon, and it would be extremely difficult to track down the issue, only to find (eventually) that the issue didn't even lie with Cinnamon. Including the dangerous property in the metadata.json file tells cinnamon-settings that it was already scanned - even though it hasn't been, so it never actually scans it. This kind of defeats the purpose of the scan in the first place.

Cobinja:
@collinss No need for explanation. I know why it was added. That doesn't mean I have to like it for this specific desklet. The only "dangerous" calls in there are spawning the settings dialog and fetching the settings file contents (plus, it messes with my git status).

JessW:
@Cobinja Hi there!  My Quote of the Day desklet also gets flagged as dangerous for using get_file_contents_utf8_sync , have you looked in to the newer async function cinnamon_get_file_contents_utf8 ?  A quick search didn't turn up any examples of how exactly to call it, and I haven't sat down myself to figure it out yet, but I thought I'd mention it.

PR #28, opened 2017-04-23
 	simple-system-monitor@ariel: Add translation

eson57:
Add l10n translation support, translation template and Swedish translation


NikoKrause:
notify author: @arielandrade

@eson57 
You changed one line to "ifc", but that seems to be nowhere defined?

eson57:
Wow... there is more errors, and I have no idea how that happened. Apparently my editor have gone crazy or something. I knew it had some bugs, but not this bad. I will update files... and throw that Notepadqq out the window.

PR #29, opened 2017-04-23
Resize icons to make it square

MrElectroNick:


collinss:
Please use a separate pull request for each applet

PR #49, opened 2017-04-26
Add translation support for Temperature

giwhub:
@swalladge


swalladge:
@giwhub nice! I didn't even realize this is where cinnamon desklets when after uploading to cinnamon-spices.

Just wondering, do you use this temperature desklet? I'd just be interested to know if someone else has found a use for it, since it only works with https://github.com/swalladge/rpi-temperature-server

giwhub:
I am not using yet,I didn't figure it out how to use it.

swalladge:
@giwhub ah no worries. Yeah there isn't much documentation with it since it was an example application to use the temperature server that was developed as a uni project.

Issue #37, opened 2017-04-25
battery@schorschii: more than 100%


NikoKrause:
Look at the screenshot. I'm using Linux Mint 17.3 Cinnamon 2.8.8

@schorschii 

![screenshot-area-2017-04-25-165640](https://cloud.githubusercontent.com/assets/8415124/25392202/4e9a6382-29d8-11e7-8d74-3843849ea88a.png)

NikoKrause:
@schorschii 
And another one:
![screenshot-area-2017-04-25-165640](https://cloud.githubusercontent.com/assets/8415124/25392285/92489c48-29d8-11e7-83f9-e79692099c6b.png)

schorschii:
Could you please provide the content of /sys/class/power_supply/BAT1/capacity? (BAT1 could also be BAT0 or CMB1)
Are you using the newest version (released 3 days ago)?

NikoKrause:
I refreshed the list. It's the newest version. It only happens, when the battery is 100%.

It's BAT0 and the content is `448`.

schorschii:
Thanks. I guess that **charge_now** is higher than **charge_full_design** and **capacity** is calculated by these values. So I have to calculate the capacity myself using **charge_now** and **charge_full**.
Could you please provide all these values when the battery is at 100%?

NikoKrause:
How do I do this?

schorschii:
The mentioned files are also in this directory: /sys/class/power_supply/BAT0 beside the capacity file

NikoKrause:
charge_now: 8739000
charge_full: 1948000
charge_full_design: 8739000
capacity: 448

schorschii:
Great. I'm going to release a fix for that soon.

NikoKrause:
@schorschii 
A quick sidenote. I just translate your desklet. There's this line in the settings-schema
```
"description": "Settings for battery@schorschii   (Version 2.4)"
```
That's a nightmare for a translator. The problem is, that if you change the version number, it's a whole new string to translate. Over and over again.

I'd recommend to add
```
"version": "2.4"
```
in the metadata.json instead. It'll be shown in the "About..." Dialog

schorschii:
Ooops. Didn't know that. I'll improve that.

NikoKrause:
Thanks! It works now.

PR #54, opened 2017-04-26
battery@schorschii: Version 2.5

schorschii:
- fix capacity indicator when battery reports capacity above 100%
- removed version from settings-schema.json and added to metedata.json


schorschii:
closes #37

PR #61, opened 2017-04-28
simple-system-monitor@ariel: update sv.po

eson57:


eson57:
Sorry for the inconvenience.

PR #79, opened 2017-04-30
Create pt_br.po

ArthurGorgonio:


NikoKrause:
@ArthurGorgonio 
The proper language ID is pt_BR.po

PR #82, opened 2017-04-30
[analog-clock@cobinja.de] Update translations

Cobinja:
This reverts a change to the pot-file, also renames the desklet to "CobiAnalogClock" in cinnamon settings.
@muzena Please don't change the pot-file for this desklet. Just use it as is. Leaving out some strings was intentional because they are already translated in Cinnamon itself.


muzena:
Ok, I think that "name" and "description" in metadata.json have not been translated, so I have add them in pot file.

Cobinja:
Not only those, but also others that don't need translation any more.

PR #83, opened 2017-05-01
File name in CommandOfTheDay has been changed to pt_BR

ArthurGorgonio:


collinss:
I not really an expert on the translations system, but wouldn't it be better to just use pt? That way it would work in pt, pt_BR, and pt_PT. @NikoKrause correct me if I'm wrong...

ArthurGorgonio:
@collinss, in both languages have some words that have differences in the translations.

NikoKrause:
@collinss
I like to stick to the language IDs, which launchpad uses: https://translations.launchpad.net/+languages

Launchpad knows Portuguese (pt) and Brazilian Portuguese (pt_BR), but it has no pt_PT.

Also it just uses fr, ru, es, but not fr_FR, ru_RU, es_ES etc.
That's currently a list of "known" language IDs, which I use for my scripts to create the translation status tables. If it finds ab unknown ID, I check if it's really a knew language or if it some of the cases like fr_FR. So we don't have fragmentation.

In this case Brazilian Portuguese is a little bit like American English compared to British English.

NikoKrause:
https://github.com/NikoKrause/cinnamon-spices-applets/blob/applets-translation-status/.translation-status/status-scripts/Language-IDs.txt

collinss:
@NikoKrause Ok, good to know. I knew there was a reason we were changing some language ids.

PR #89, opened 2017-05-01
bbcwx@oak-wood.co.uk: remove redundant en.po file

NikoKrause:
@tipichris


tipichris:
Working from memory a bit here, but I don't think en.po is redundant. This was because the degree symbols are not ASCII, and the translation template file (for reasons that I can't remember) had to be ASCII. So some of the translations in en.po, particularly those including degree symbols, differ from the translation template.

tipichris:
So check entries like msgid "%fF", msgid "%fC"

tipichris:
@NikoKrause thought I'd better throw in a mention to draw your attention to my comments about en.po not being redundant.

NikoKrause:
Oh sorry, I did  a quick overview and all the translations seemed the same.
Isn't it possible to adapt the template .pot file?

NikoKrause:
@tipichris 
I mean adding the ° symbol in the code?

tipichris:
@NikoKrause There was a reason why I couldn't use non-ascii characters in the msgid, but I can now remember whether it was an inherent limitation in gettext or just in one of the tools I used to manage translations. Either way, I had to adopt this strategy early on and changing now would mean changing all the .po files to match, and possibly throw the launchpad translation page into confusion.

tipichris:
@NikoKrause perhaps related to one of these:
https://www.gnu.org/software/gettext/FAQ.html#nonascii_strings
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=594059

NikoKrause:
@tipichris 
Sorry, I didn't wanted to create any circumstances.
I'll revert this change then. Is that okay?

tipichris:
That would be great. Thanks.

PR #101, opened 2017-05-02
Refresh when second next changes

tipichris:
When not smoothing the seconds, having the desklet refresh every 1000 ms can lead to a situation where the second consistently updates 999ms into the second, so the clock appears to be nearly 1s slow. This patch causes the next refresh to be immediately after the next change of second.


tipichris:
@schorschii FYI

schorschii:
Thanks, looks good!

PR #103, opened 2017-05-04
clock@schorschii: Always smooth hour hand

tipichris:
The hour hand of an analogue clock should always indicate partial elapsed hours, even when seconds and minutes are not 'smoothed'. At 9:55 the hour hand should not point at 270° straight to the nine. We are so used to this that if it is not smoothed then close to the hour it easy to glance at the clock and see it as 8:55. I've checked all the clocks I can find, both real and virtual. Smoothing the hour is always done. 

I've left the minute as this is less universal. Some do, some don't. I think I prefer not, but I'm not sure. Perhaps smoothing minutes should be an additional configuration option?

@schorschii

### DIFF
--- a/clock@schorschii/files/clock@schorschii/desklet.js
+++ b/clock@schorschii/files/clock@schorschii/desklet.js
@@ -93,15 +93,13 @@ MyDesklet.prototype = {
 		let mseconds = this._displayTime.get_microsecond();
 
 		// calc pointer rotation angles
-		let hours_deg = 0;
+		let hours_deg = ((hours+(minutes/60))*360/12);
 		let minutes_deg = 0;
 		let seconds_deg = 0;
 		if (this.smooth_seconds_hand == true) {
-			hours_deg = ((hours+(minutes/60))*360/12);
 			minutes_deg = ((minutes+(seconds/60))*360/60);
 			seconds_deg = ((seconds+(mseconds/1000000))*360/60);
 		} else {
-			hours_deg = (hours*360/12);
 			minutes_deg = (minutes*360/60);
 			seconds_deg = (seconds*360/60);
 		}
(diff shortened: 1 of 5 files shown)


schorschii:
@tipichris I agree with you that it's better to smooth the hour hand in every case. I'm not sure what to to with the minute hand. So, as you said, it would be the best to add an option for this. Would you do that? Otherwise I'll write it on my ToDo-List.

tipichris:
How long is your ToDo-List?!

I can do it as I think it will be quite quick. Though I don't have a lot of time in the next few weeks. **But** it will mean the translation files need some additional work. As well as having an extra preference to translate the label and tool tip for, the msgid for current preference label and tooltip will need to change to make any sense. 

Whilst I'm at it, what do you think about reducing the refresh time if show-seconds-hand is false? Seems at the very least it should never be 100ms (which noticeably increases CPU usage), but actually it could go down to something like 5s.

tipichris:
I've done the preference for smoothing the minute hand. 

My automated rebuilding of the .po files seems to have put messages in a different order, so the commit looks more frightening than it really is.
@schorschii

schorschii:
Thanks, this looks good. I'm going to update the german translation immediately after this commit is merged. With that I'm going to add some differentiations for the refresh time.

In my opinion this pull request can now be merged.

tipichris:
If it's any help, I did have a quick look at the refresh time too. What I came up with was this, which refreshes about every three seconds if second hand is not showing, and does so in a way that ensures it will refresh close on the change of second.

    let timeoutval = 100;
    if (this.show_seconds_hand == false)
        timeoutval = Math.ceil(3000 - (1000*seconds + mseconds/1000) % 3000);
    else if (this.smooth_seconds_hand == false)
        timeoutval = Math.ceil(1000 - mseconds/1000);
    this.timeout = Mainloop.timeout_add(timeoutval, Lang.bind(this, this.refresh));

schorschii:
Cool, of course this is a help. I'll include this in my next commit.

PR #145, opened 2017-06-20
clock@schorschii:Danish translation

Alan01:


NikoKrause:
The file has to be in the po folder, please change and rebase your commit

PR #170, opened 2017-07-03
simple-system-monitor: add some simple imports for gtop, nmclient and…

brownsr:
… network manager

In Cinnamon 3.4 the desklet is failing, perhaps the new cjs version,
and the try/catch way of catching missing dependencies and including
the imports is not working.
While this change does not make the dependency messaging work, it does
restore basic functionality for a user that has the required
dependencies installed


brownsr:
ping @arielandrade

arielandrade:
Thank you! I don't have write access to merge it @NikoKrause do you?

PR #189, opened 2017-09-12
xkcd@rjanja: use icon in desklet itself

NikoKrause:
this ensures that the icon is shown in the desklet settings

@rjanja are you ok with this?


rjanja:
@NikoKrause That's fine. I don't use this desklet anymore and would be happy to see it adopted by another maintainer.

NikoKrause:
OK. I'll change the author to none, so anyone can take care of this desklet.

PR #191, opened 2017-09-16
googleCalendar@javahelps.com: add translation support

NikoKrause:
... and German translation

@slgobinath fyi
If Strings in the desklet.js file need to be translated, you need to put them in brackets with an underscore, i.e. like this `_("")`

The translation template file (important for translators) is `po/googleCalendar.pot`. If you add/change translatable strings, you don't need to update the translation files `.po`, but you need only to update the template `.pot` file with the following command:
```
cd ~/.local/share/cinnamon/desklets/googleCalendar@javahelps.com
cinnamon-json-makepot --js po/googleCalendar.pot
```


slgobinath:
@NikoKrause Thank you very much for the information and your contribution :-)

PR #215, opened 2017-11-08
Add support for OpenWeatherMap with a free account

Kallys:
OpenWeatherMap's [API](https://openweathermap.org/api) differs with a free account ([5 day / 3 hour forecast](https://openweathermap.org/forecast5) or [16 day / daily forecast](https://openweathermap.org/forecast16))
This commit allow users to use the 5 day forecast API (the only one available with free accounts).


NikoKrause:
@tipichris

tipichris:
Actually there is another, more serious issue. The change breaks the original OWM driver when used with a latitude and longitude. I think this is because the variable latlon is out of scope in the new _get_apiforecasturl function. This does need addressing before changes are approved.

PR #217, opened 2017-11-18
Battery: Changed variable declarations from var to let.

icarter09:
@schorschii  Changed variable declarations from var to let. If the 'var' was intentional, then feel free to leave a comment and close this pull request.


schorschii:
@icarter09 Thanks! In my opinion this request can be merged.

PR #232, opened 2018-01-09
bbcwx@oak-wood.co.uk: Update Catalan translation

rbuj:


tipichris:
Thanks for the contribution. Unfortunately, whilst you've been working on this I updated the translation template. Could you try and merge your work with the updated template. I'm not a gettext expert, but I thing something like this should work

`msgmerge --update --backup=simple ca.po bbcwx@oak-wood.co.uk.pot`

PR #243, opened 2018-01-12
added new desklet notes@schorschii

schorschii:
- added my notes desklet


schorschii:
I have tested the desklet on two more newly installed Mint 18.3 computers and I currently can't reproduce the bug that you have mentioned when you drag it with the mouse. Does this also happen with my other desklets on your system?
Per default, new instances of my desklet are displaying the content of "~/note.txt", as long as you don't choose another file in settings. Is this behavior fine or should i change this?

jaszhix:
I retested in another VM and it seems to be fine, sorry for the false negative.

PR #247, opened 2018-01-29
Toward a better wind direction approximation (bbcwx@oak-wood.co.uk)

Kallys:
As @tipichris mentionned in #215 , wind direction was poorly estimated.
I see that others computation issues are already fixed, but this one ran through my mind...
How to average angles?
I choose a naive solution: get cartesian coordinates, average them, and then get angle back.

I'm not sure if this really need a PR, but please tell me if this method leads to better accuracy! (I can't access day forecast since I have no 16 day API account).


Kallys:
Ok, thanks for those details!
We should probably get more information about the "alignment of 3 hours slices" but I'm not sure this is really mandatory.
So I'm gonna remove this PR, thanks a lot!

PR #262, opened 2018-02-17
notes@schorschii: Make it possible not to have a background

getkey:
Hello @schorschii, thanks for the great desklet!

I made this modification because your desklet looks better with my desktop theme without a background. Feel free to suggest better ways to do this if it feels to hacky. 😉


schorschii:
@getkey That's a good idea, thanks! In my opinion it can be merged.

PR #263, opened 2018-02-18
googleCalendar@javahelps.com ru.po update

Ilis:
ru.po update


Ilis:
I've updated ru_po, but noticed that string for shortened month name May is missed in pot.

PR #300, opened 2018-05-03
yfquotes@thegli: new/old desklet for stock quotes information

thegli:
based on stocks@fthuin, but with fixed/working data retrieval from Yahoo Finance, and various minor improvements


yfquotes@thegli/files/yfquotes@thegli/desklet.js
```diff
@@ -0,0 +1,261 @@
...
+var dirPath = "yfquotes@thegli";
+var deskletDir = GLib.get_home_dir() + "/.local/share/cinnamon/desklets/" + dirPath;
+
+var YahooQueryStockQuoteReader = function () {
+};
+
+YahooQueryStockQuoteReader.prototype = {
+	constructor: YahooQueryStockQuoteReader,
+	yahooQueryBaseUrl: "https://query1.finance.yahoo.com/v7/finance/quote?symbols=",
+	getStockQuotes(quoteSymbols) {
```
jaszhix: ES6 method definitions only work 100% in Cinnamon 3.8, based on mozjs52. mozjs38 also has this, but some people (usually Debian users) run Cinnamon 3.4-3.6 with CJS 3.2, based on mozjs24. You can specify compatibility in by adding `"cinnamon-version": ["3.8"]` to metadata.json.

thegli: Thank you for the feedback.
I can revert the method definition changes. Would it then run properly with CJS 3.2 and mozjs-24? I have currently no such setup at hand to test myself.
And does the compatibility flag mean "works up to that Cinnamon version"?

jaszhix: No problem - I would have to look more closely. Setting `cinnamon-version` that way would restrict it to being installed on Cinnamon 3.8 only. There's also another option you can use, `js-version`, you could also try adding `"js-version": ["3.4.0", "3.6.0", "3.8.0"]` to metadata.json. No xlets use it AFAIK and I haven't tested it, so can't vouch that it works reliably. If you do go that route I can test in VM to make sure its working.

thegli: I decided to undo the ES6 notation. Tested it successfully with a Debian 9 VM (Cinnamon 3.2). Also fixed the error log flooding.

Issue #268, opened 2018-02-26
[analog-clock@cobinja.de] Clock ignores Timezone setting and just shows local time.


kauffy:
```
 * Desklet version/Build date (to be found in "About.." or in "metadata.json")
 * Cinnamon version (cinnamon --version)
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...)
Mint 18.3 Sylvia

 * Graphics hardware *and* driver used
Intel Integrated-- non-proprietary driver

 * Snippet or attachment of ~/.cinnamon/glass.log
Nothing interesting in here.
info t=2018-02-26T03:24:55.280Z Loaded desklet analog-clock@cobinja.de in 116 ms
info t=2018-02-26T03:26:46.560Z Loaded desklet analog-clock@cobinja.de in 104 ms


 * Snippet or attachment of ~/.xsession-errors
Cjs-Message: JS WARNING: [/home/fletcher/.local/share/cinnamon/desklets/analog-clock@cobinja.de/3.0/desklet.js 165]: Too many arguments to method Gio.File.delete: expected 1, got 2
Cjs-Message: JS WARNING: [/home/fletcher/.local/share/cinnamon/desklets/analog-clock@cobinja.de/3.0/desklet.js 167]: Too many arguments to method Gio.File.delete: expected 1, got 2
Cjs-Message: JS WARNING: [/home/fletcher/.local/share/cinnamon/desklets/analog-clock@cobinja.de/3.0/desklet.js 169]: Too many arguments to method Gio.File.delete: expected 1, got 2
Cjs-Message: JS WARNING: [/home/fletcher/.local/share/cinnamon/desklets/analog-clock@cobinja.de/3.0/desklet.js 157]: Too many arguments to function GLib.file_set_contents: expected 2, got 3
Cjs-Message: JS WARNING: [/home/fletcher/.local/share/cinnamon/desklets/analog-clock@cobinja.de/3.0/desklet.js 28]: Too many arguments to function GdkPixbuf.get_file_info: expected 1, got 3
Cjs-Message: JS WARNING: [/home/fletcher/.local/share/cinnamon/desklets/analog-clock@cobinja.de/3.0/desklet.js 28]: Too many arguments to function GdkPixbuf.get_file_info: expected 1, got 3
[...]

PR #305, opened 2018-05-13
Added fix for clock ignoring timezone changes.

icarter09:
@Cobinja: This PR is in reference to issue #268.
I wasn't for sure if the code block was needed that I removed. The problem with the issue was that "_initialUpdate" never gets reset so the clock doesn't properly update when a user changes the timezone. If the conditionals that I removed are needed, then two other possible fixes would be:

1) Set "_initialUpdate" to true in the "_onTimezoneChanged" function.
2) Create a variable named "_hasTimezoneChanged", set it to true on "_onTimezoneChanged" function, add that to the conditional statements I removed (lines 451 & 452), and then set it to false once the hours and minute on the clock are update.

### DIFF
--- a/analog-clock@cobinja.de/files/analog-clock@cobinja.de/3.0/desklet.js
+++ b/analog-clock@cobinja.de/files/analog-clock@cobinja.de/3.0/desklet.js
@@ -448,13 +448,10 @@ CobiAnalogClock.prototype = {
     let minutes = this._displayTime.get_minute() % 60;
     let seconds = this._displayTime.get_second() % 60;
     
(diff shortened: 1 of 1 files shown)


Cobinja:
Soryy, I have to reject this.
This would mean that the desklet updates the hour clock every second, which is unecessary rendering load.
My PR is on its way.

icarter09:
@Cobinja no worries! I wasn't for sure on which fix to implement so I pushed the one and referenced the other two ideas as possible fixes. I knew you would know which was the best solution. Thanks!

PR #306, opened 2018-05-14
analog-clock@cobinja.de 2 bugfixes

Cobinja:
- Fix tz clock update after timezone change (Fixes https://github.com/linuxmint/cinnamon-spices-desklets/issues/268)
- Fix file handling for newer GLib/GdkPixbuf versions


kauffy:
This appears to be working now! The time is correct the moment the desklets appear, per their timezone setting.

PR #323, opened 2018-07-21
Update cpuload@kimse support HiDPI scaling; fixes #322

### LINKED ISSUE #322: [cpuload@kimse] HiDPI display issue.
```
 * Desklet version 0.3
 * Cinnamon version 3.8.8
 * Distribution - Mint 19
 ```

**Notify author of desklet**

@kimse

**Issue**

Fonts are overlapping on HiDPI displays.

Redsandro:
The problem with using font size for location calculations is that font "points" are automatically scaled with the HiDPI scaling settings in CSS. But javascript position "pixel" calculations are **not**.

Also, half point fonts don't render nicely. I've increased the font size a tiny little bit.

### DIFF
--- a/cpuload@kimse/files/cpuload@kimse/desklet.js
+++ b/cpuload@kimse/files/cpuload@kimse/desklet.js
@@ -36,8 +36,8 @@ CpuusageDesklet.prototype = {
     setupUI() {
 
         this.minDeskletWidth = 175;
-        this.largeFontSize = 20;
-        this.normalFontSize = 13;
+        this.largeFontSize = 22;
+        this.normalFontSize = 14;
 
         // Create a main window        
         this.window = new Clutter.Actor();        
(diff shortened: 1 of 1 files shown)


brownsr:
@ipimpat

ipimpat:
I tested it, but the position of the text is not right on my screen (2x 1920x1080)

hidpi

thegergo02:
I don't know what value the `global.ui.scale` is holding, but I think check if scaling is applied, and if not don't use this fix, but if scaling is applied use the `global.ui.scale` value.

Redsandro:
@ipimpat sorry I couldn't figure out how to do the proper calculations. Someone more knowledgeable than I should look at the PR (or perhaps it's fixed in a different one). Either way, not using the scale factor causes a problem at higher scales, because the text is unreadibly micro sized. I stopped using this extension for now.

ipimpat:
@Redsandro if you wan't to, we can work together on the issue. 

Drop me an e-mail, if you are interested.

Issue #316, opened 2018-06-13
top@ryannerd PID in top sometimes has wrong PID [HELP WANTED]


RyanNerd:
```
 * Desklet version/Build date (1528658206)
 * Cinnamon version (3.6.7)
 * Distribution - (Mint 18.3)
 ```

**Issue**
I am using the Top parser from [here](https://github.com/devalexqt/topparser)
On occasion the PID is dropping the first digit. I'm having trouble figuring it out since the code is using regular expressions (something I'm not very good at). The code that parses the PID is [here](https://github.com/linuxmint/cinnamon-spices-desklets/blob/376e54a6970508366aa622f351d266f2aab9a2cd/top%40ryannerd/files/top%40ryannerd/desklet.js#L81)

I'm hoping someone who is really good at regular expressions can help me out.

**Steps to reproduce**
Run top as a desklet and sometimes the PID will not match the PID if you run `top` in the command line.

**Expected behaviour**
PID shouldn't drop a digit from the display.

**Other information**
Posted by the author of this desklet requesting help from the LM community.

icarter09:
@jaszhix I can provide some help with this issue.

@RyanNerd I ran the Top desklet for quite awhile and compared it against running "top" in the terminal. I didn't notice the Top desklet dropping any digits from the PID. Can you provide a screen capture of the PID dropping a digit or can you provide the CMD that is associated with the PID that is getting dropped?

icarter09:
@RyanNerd thanks for the screen captures! I can see the problem you are talking about. I'll take a look into the code and the regex parsing tonight.

icarter09:
@RyanNerd I might have found the problem, but not for sure. I did some researching into PID numbers and how they are assigned/created. One problem is that my machine is not generating 5 digit PID numbers, but when I did statically set the value to be 5 digits, it still worked just fine. The values were not cut off and were displayed correctly, unlike what your screen captures show. 

I put some comments in the PR for you as well.

If this doesn't work, just let me know and I'll continue working with you to get it fixed.

RyanNerd:
@icarter09 Thank you for the PR.
Unfortunately it did not solve the issue.
Fortunately it helped me identify [where the issue is](https://github.com/linuxmint/cinnamon-spices-desklets/blob/968a741cb36ac846101173b5060dab2e7c966856/top%40ryannerd/files/top%40ryannerd/desklet.js#L161) and I now know how to fix it.

[...]

PR #328, opened 2018-07-25
Removed parseInt logic for PID issue fix.

icarter09:
@RyanNerd this PR is for issue #316.

I'm not 100% sure if this will fix the issue. All the regex stuff that the desklet is doing checks out ok and also does not have anything to do with the PID process numbers being displayed. I debugged several spots in the code and believe that this is the one location that could/might be responsible for the cutting off of the digits. Give this PR a shot and see if it helps out. If not, just let me know and I'll go back to the drawing board.


icarter09:
@jaszhix can you close/delete this PR? It is no longer needed.

PR #339, opened 2018-09-19
analog-clock@cobinja.de: Avoid passing context to SignalManager

jaszhix:
Passing context to the SignalManager constructor creates circular references, which makes memory management more complicated for the JS engine.

This is a fix for https://github.com/linuxmint/Cinnamon/pull/7868.

@Cobinja


Cobinja:
First glance looks good.
I'll test this tonight after work before confirming

PR #341, opened 2018-10-13
add sharkle@lufinkey

lufinkey:
Added Sharkle from the game [Night in the Woods](https://www.youtube.com/watch?v=O2qRVwQEOh8)


jaszhix:
Hi @lufinkey, this is pretty neat! 

I have one concern though, which is its always animating in the background causing the Cinnamon thread to use 20%+ CPU usage on idle. Maybe the animation could be suspended when the cursor is not hovering over the desklet. You could also check if the desklet overlay is visible (`Main.deskletContainer.isModal` on 3.8+).

sharkle@lufinkey/files/sharkle@lufinkey/desklet.js
```diff
@@ -0,0 +1,285 @@
...
+		if(this.animation!=null)
+		{
+			Mainloop.source_remove(this.loop_id);
+			this.loop_id = 0;
+		}
+		this.frame = 0;
+		this.animation = animation;
+		if(this.animation!=null)
+		{
+			this.loop_id = Mainloop.timeout_add(1000.0/this.animation.fps, Lang.bind(this, this._update_loop));
```
jaszhix, line 55: When removing the desklet the looping doesn't end. These errors are seen continuously, note these can only be seen using cjs master:

```
(cinnamon:18866): Cjs-CRITICAL **: 02:45:00.794: Object St.Bin (0x55a3a4d4a3b0), has been already deallocated - impossible to access to it. This might be caused by the fact that the object has been destroyed from C code using something such as destroy(), dispose(), or remove() vfuncs
== Stack trace for context 0x55a3a2cc64c0 ==
#0 0x55a3a3604c48 i   /usr/share/cinnamon/js/misc/fileUtils.js line 210 > Function:107 (0x7f92437694d8 @ 30)
#1 0x55a3a3604b90 i   /usr/share/cinnamon/js/misc/fileUtils.js line 210 > Function:91 (0x7f92437699a0 @ 237)
#2 0x55a3a3604ad0 i   self-hosted:914 (0x7f92ddee3ef0 @ 346)

(cinnamon:18866): Cjs-CRITICAL **: 02:45:00.794: Object St.Bin (0x55a3a4d4a3b0), has been already deallocated - impossible to access to it. This might be caused by the fact that the object has been destroyed from C code using something such as destroy(), dispose(), or remove() vfuncs
== Stack trace for context 0x55a3a2cc64c0 ==
#0 0x55a3a3604c48 i   /usr/share/cinnamon/js/misc/fileUtils.js line 210 > Function:112 (0x7f92437694d8 @ 86)
#1 0x55a3a3604b90 i   /usr/share/cinnamon/js/misc/fileUtils.js line 210 > Function:91 (0x7f92437699a0 @ 237)
#2 0x55a3a3604ad0 i   self-hosted:914 (0x7f92ddee3ef0 @ 346)

```

You could use a function [like this](https://github.com/linuxmint/cinnamon-spices-applets/blob/64c6130057a2ae2dce4e0fcc4b663e997ae8b3e3/IcingTaskManager%40json/files/IcingTaskManager%40json/3.8/utils.js#L189) to check if an object is finalized, but stopping the loop in `on_desklet_removed` should be enough.

lufinkey:
Gotcha I'll address those issues. I'm pretty certain that I did animation in a less than optimal way though. There wasn't much documentation to go off of. I'm basically storing each image in a view or "bin" and just cycling through them. Is there a better way to do this?

jaszhix:
[This site](http://devdocs.baznga.org/) is a good resource for JS binding-specific docs. There's probably a way to do it with Clutter, but never looked into it myself. Cinnamon uses the Tween module in cjs, but not sure I've seen it used to compose an animation with image frames.

brownsr:
ping @lufinkey

lufinkey:
@brownsr hey sorry! I haven't looked at this in a while. I'll try to get this working soon!

PR #365, opened 2018-12-15
netusage@30yavash.com v1.0.3 Option to pick vnstati output with multiple instances

pdcurtis:
* Added desklet_id to various functions
  * Add Configure (settings-schema.json) to applet
  * Provide options to choose different vnstati formats including a user specified format
  * Tidy code to remove trailing spaces
  * Change Icon to be unique and have better affordance
  * Add CHANGELOG.md and Update README.md

For information - these are similar changes to those made on applets vnstat@cinnamon.com and netusagemonitor@pdcurtis with added ability to have several different vnstati displays can be on the desktop


pdcurtis:
* set ` "prevent-decorations": true`  in `metadata.json`

Force pushed change to minimize number of commits

PR #366, opened 2018-12-18
netusage@30yavash.com v1.0.4 Extend choice of display formats

pdcurtis:
* Extend number of choices of vnstati formats
  * Correct Icon for Cinnamon 4.0
  * Add CHANGELOG.md
  * Update README.md


netusage@30yavash.com/files/netusage@30yavash.com/desklet.js
```diff
@@ -152,26 +152,18 @@ MyDesklet.prototype = {
...
-                         }
-
                          if(this.extendedDisplay == "userDefined" ) {
                                if (this.useVnstatiCommandString) {
                                      GLib.spawn_command_line_sync('vnstati ' + this.vnstatiCommandString + ' -ne -i ' + this._device + ' -o /tmp/vnstatlmapplet.png');
                                } else {
                                      GLib.spawn_command_line_sync('vnstati -s -ne -i ' + this._device + ' -o /tmp/vnstatlmapplet.png');
                                }
+                         } else {
+                                 GLib.spawn_command_line_sync('vnstati ' + this.extendedDisplay + ' -ne -i ' + this._device + ' -o /tmp/vnstatlmapplet.png');
```
jaszhix: It might be better to try using `Util.spawnCommandLineAsync` or [spawn_async](https://github.com/linuxmint/Cinnamon/blob/bbbf728a01845a151f9da4d50a74804f2af53af7/js/misc/util.js#L91) here instead, so if the process misbehaves it doesn't block the main thread.

netusage@30yavash.com/files/netusage@30yavash.com/settings-schema.json
```diff
@@ -4,41 +4,44 @@
...
+        "tooltip": "This feature offers the Classic (original) display with a number of other options including use of a user defined string for vnstati when enabled below"
    },
 
     "useVnstatiCommandString": {
         "dependency": "useExtendedDisplay",
         "type": "checkbox",
         "default": false,
         "description": "Enable Custom Display Option",
-        "tooltip": "Check box"
+        "tooltip": "Beware - no error checking - if you crash the applet try restarting Cinnamon"
```
jaszhix: I recommend handling the errors or sanitizing user input instead of instructing people to restart Cinnamon if the desklet crashes.

pdcurtis: This was somewhat tongue in cheek as it is impossible to know what a program such as vnstati will do with faulty input and I did once upset vnstati with an attempt to use one of the parameters which replaced the vnstati image header with a text string which tripped the the try...catch

There are two choices:
  * Remove the option completely.
  * Make the the message even stronger and recommend backing up first. 

I favour removing it as I have now added all the reasonable display type options and anything else has a poor benefit/risk ratio . The dangers were the reason why there was an extra checkbox to activate it although I found it very useful during testing.

I will push an extra commit to remove the option completely.

netusage@30yavash.com/files/netusage@30yavash.com/desklet.js
pdcurtis: @jaszhix I have a sense of deja vu on this as I originally suggested the same to Clem who wrote vnstat@cinnamon.com which this is based on. The problem there was that `vnstat` had to complete before the picture which `vnstati` generates can be displayed so it could sometimes be picking up an image which was days out of date and we eventually had to revert the change. 

Here the problem is similar in that I only use a single temp file for all the instances and if one does it asynchronously one picks up an  image from a random instance which is very confusing as they all keep changing! 

If you can suggest a simple way of defining a path to a different image for each instance I am very happy to try an async version but, as I am away over Christmas, I would prefer to leave it until the next version.

pdcurtis:
Sorry about the force push but there was an error in the update and I did not want multiple commits

netusage@30yavash.com/files/netusage@30yavash.com/desklet.js
jaszhix: I looked at the code in more detail, and this is the solution I came up with:

```js
_updateGraph: function() {
        let path = `/tmp/${this._uuid}_${this.instance_id}`;
        let instancePath = Gio.File.new_for_path(path);
        if (!instancePath.query_exists(null)) {
            if (!instancePath.make_directory(null)) {
                global.log(this._uuid, `Cannot make directory: ${path}`);
                return;
            }
        }
        if (this._device != "null") {
            let command;
            if (this.useExtendedDisplay) {
                command = 'vnstati ' + this.extendedDisplay + ' -ne -i ' + this._device + ' -o ' + path + '/vnstatlmapplet.png';
            } else {
                command = 'vnstati -s -ne -i ' + this._device + ' -o ' + path + '/vnstatlmapplet.png'
            }
            try {
                Util.spawnCommandLineAsync(command, () => {
                    let l = new Clutter.BinLayout();
                    let b = new Clutter.Box();
                    let c = new Clutter.Texture({keep_aspect_ratio: true, filter_quality: 2, filename: path + '/vnstatlmapplet.png'});
                    b.set_layout_manager(l);
                    b.add_actor(c);
                   // this.imageWidget.destroy_all_children();
                    this.imageWidget.set_child(b);
                });
            } catch (e) {
                this.warnings = new St.BoxLayout({vertical: true});
                this.missingDependencies = new St.Label({text: _("Please make sure vnstat and vnstati are installed and that the vnstat daemon is running!")
                                        + "\n" + _("In Linux Mint, you can simply run 'apt install vnstati' and that will take care of everything.")
                                        + "\n" + _("In other distributions it might depend on the way things are packaged but its likely to be similar.")
                                        + "\n" + _("The Interface detected was: " + this._device )
                });
                this.warnings.add(this.missingDependencies);
                this.setContent(this.warnings);
            }
        }
    }
```

Its tested and working. Note, I'm not using Network Manager and had to comment out `_updateDevice`, and put my real network device in metadata.json.

Edited the example a few times because the error callback wasn't getting called (this is fixed in https://github.com/linuxmint/Cinnamon/pull/8130).

jaszhix: One other thing I noticed was this might be causing a slow memory leak when `imageWidget` has new children set. You might also want to add

```js
this.imageWidget.destroy_all_children();
```
before
```js
this.imageWidget.set_child(b);
```

pdcurtis:
@jaszhix Many thanks yet again.It will take me a few days as I have no access to git at present.. 

I will also need to check on my applets as the code which is leaking may also be present. 

I also need to handle condition when no network better

pdcurtis:
@jaszhix I have had a good look through the code although I still have not been able to run it and I have a couple of quick questions.
Are there any restriction on the versions of cinnamon which are needed to run it in particular the code you had problems with checking/creating the folder and spawnCommandLineAsyn()
Is there a reason to create the new paths rather than different names for the images which could get round the problem?
Thanks again for the work and testing

jaszhix:
No problem. You're right, the code could probably be simplified a bit by just putting the instance ID in the image file name. `spawnCommandLineAsync` has been in Cinnamon since 3.6. The way I called it in the example should be compatible with the PR that I linked earlier as well.

pdcurtis:
@jaszhix I have worked on the basis of the minimum number of changes required so I have used the instance-id in the filenames.
The Command line calls are now asynchronous but not using `spawnCommandLineAsync` as that you noted it is limited to >3.4
I have added the code to prevent the potential memory leak.

I have noticed two things in testing.
  * `max-instances` in `metadata.json` does not seem to be respected in any of the desklets I have tested.
  * I have had to put a small delay in initiating the desklet as instance_id is not immediately available and returns `undefined` as part of the filename.

Have you seen either of these anomalies elsewhere?

I am otherwise happy for this to be merged  and have removed the WIP - you may wish to squash the three commits or I can do so and force push it up.

netusage@30yavash.com/files/netusage@30yavash.com/4.0/desklet.js
```diff
@@ -129,20 +129,22 @@ MyDesklet.prototype = {
...
         try {
+           let image = `/tmp/vnstatImage${this.instance_id}.png`;
            if (this._device != "null") {
                    if (this.useExtendedDisplay) {
-                           GLib.spawn_command_line_sync('vnstati ' + this.extendedDisplay + ' -ne -i ' + this._device + ' -o /tmp/vnstatlmapplet.png');
+                           GLib.spawn_command_line_async('vnstati ' + this.extendedDisplay + ' -ne -i ' + this._device + ' -o ' + image );
 
                    } else {
-                        GLib.spawn_command_line_sync('vnstati -s -ne -i ' + this._device + ' -o /tmp/vnstatlmapplet.png');
+                       GLib.spawn_command_line_async('vnstati -s -ne -i ' + this._device + ' -o ' + image );
```
jaszhix: This is run asynchronously, but the image fetching is still synchronous. In the example I provided, the image is added to the texture inside the callback, which is when the command is finished running. So in this case the image could be added to the texture and not yet exist.

pdcurtis: @jaszhix  Yes I realise which is why the images are separated by instance so the worst effect is using an image created a few seconds earlier which is insignificant when vnstat only updates every 5 minutes. Your solution is obviously more elegant and could be included for the 4.0 but the basic code is good back to 2.2.

I was actually thinking of slowing the loop dramatically and splitting into say a one second and a 30 second part so there was no visible delay after a configuration change yet reduce the processor load. I have already done that in another applet when too frequent reads of the configuration caused segfaults (That issue was cleared long ago before you worry).

Its the old problem of the best being the enemy of good enough! 

My other future enhancement is an advanced option too allow the interface to be specified which is desirable for interfaces not controlled by the network manager and also VPN tunnelling both of which need a user experienced enough to set up vnstat for interfaces not automatically identified at install time.

Any comments on max-instances and the need for the delay?

jaszhix: Max instances not working is a Cinnamon bug. When I was testing my example, I didn't need a delay. I haven't tested this new commit yet, but I don't think this is a working async implementation.

I think supporting 3.6+ is reasonable. Mint 18.x users can upgrade to 18.3. The only users left out is Debian users stuck on 3.2, and that's a problem with Debian IMO. Nobody should be using mozjs24 anyway, it's from 2013. There's no gain to supporting ancient environments, especially when it holds back code quality.

pdcurtis: Working fine for me with 5 instances.

I have always had the principle of supporting all LTS versions otherwise why have them!

jaszhix: I'm not trying to get you to write code a certain way, I'm just looking at this and seeing that it's not working as it should. If you want to use GLib, fine by me - but use it correctly. This method doesn't have a callback, so you are doing guess work. You could also copy [this function](https://github.com/linuxmint/cinnamon-spices-applets/blob/master/IcingWindowSaver%40json/files/IcingWindowSaver%40json/3.6/applet.js#L20) to your files and use it instead, and not worry about compatibility.

PR #374, opened 2018-12-28
eKreta@thegergo02: Display your grades from the eKreta API

thegergo02:


jaszhix:
Hi @thegergo02, I don't have a way to test the API requests since this seems to be a closed education-based service in Hungary, but did notice this warning:

```
Cjs-Message: 09:28:26.120: JS WARNING: [/usr/share/cinnamon/js/misc/fileUtils.js line 294 > Function 107]: Too many arguments to method Soup.Message.set_request: expected 3, got 4
```

From [GJS binding docs](http://devdocs.baznga.org/):

> set_request(content_type, req_use, req_body)
>
> Parameters:
> - content_type (String) — MIME Content-Type of the body
> - req_use (Soup.MemoryUse) — a Soup.MemoryUse describing how to handle req_body
> - req_body (ByteArray) — a data buffer containing the body of the message request.

thegergo02:
I will fix it, somewhere I read 4 arguments. How can you see this log? I use the looking glass and Developer Tools, and that doesn't show warnings like this.

jaszhix:
If you're using LightDM, it should be in ~/.xsession-errors, otherwise you might need to run `cinnamon --replace` to get stdout.

thegergo02:
I see, thanks!

PR #391, opened 2019-01-29
InternetTimeDesklet@stefan : new

stefan12O:
I would like to publish this desklet.


claudiux:
Hi @stefan12O,

I'm using your nice desklet (and I just have added a Star on its page).

You did say that your biggest challenge is handling with GitHub. As me a few months ago. So, I offer you a method to propose clean PRs: https://github.com/claudiux/docs/blob/master/HOW_TO_MAKE_A_CLEAN_REFORK.md

stefan12O:
Hi @claudiux,
thank you very much. Good to know, that I am not the only one. :-)
I will go through the document these days. At a first glance, I have to say, I should have discovered it before.
Thank you for using and liking.

PR #413, opened 2019-03-21
clockTow@armandobs14: date format in user language

Hukuta:
I suggest using the date format and language depending on the user's locale (LANG env variable).

How it will be: 
date_ar
arabic

date_en
english (us)

date_ja
japanese

etc.


Hukuta:
@armandobs14, agree with the change?

armandobs14:
@Hukuta, it sounds good.

PR #436, opened 2019-07-06
show-remote-ip-desklet@nejdetckenobi: Adds a desklet that shows your remote IP.

nejdetckenobi:


nejdetckenobi:
QA check is wrong. 

- I can not camelcase the functions offered by it.
- "imports" is not an undeclared variable in a desklet.
- I am not sure if it will work when I use the method shorthand in ES6

brownsr:
Sorry, I can't merge this with the reference to an obscure website in there, although it looks harmless enough now, I can't guarantee that something more malicious might be put there.

Regards

Simon

leigh123linux:
That site has been used for many years.

https://major.io/icanhazip-com-faq/

brownsr:
OK !  Thanks @leigh123linux .

Issue #279, opened 2018-03-03
simple-system-monitor@ariel and netusage@30yavash.com need to be updated to support new NM library


coderforlife:
Distribution: Fedora 27
Author of `simple-system-monitor`: @arielandrade 
`netusage` is unclaimed

Both of these desklets cannot be loaded on Fedora 27 (and possibly future versions of Linux Mint or other distros) due to the move from `NMClient` and `NetworkManager` libraries to `NM`. Attempting to add this applet will cause Cinnamon to continually crash on Fedora 27.

More about the issue is discussed in https://github.com/linuxmint/cinnamon-spices-applets/issues/1616 which talks about it for the applet multicore-sys-monitor@ccadeptic23 and has a fix in PR https://github.com/linuxmint/cinnamon-spices-applets/pull/1646. A similar fix would be required for each of these desklets as well.

This may be related to #223 which mentions the desklet continually crashing on Fedora 26.

coderforlife:
Due to messed up formatting initially I am unsure if the mention of @arielandrade went through so repeating it.

arielandrade:
Hello Jeffrey, I'm sorry but I can't mantain the desklet right now, I'll
donate it so somebody else might fix it.

pdcurtis:
I fixed the same issue in the matching applet vnstat@linuxmint.com to the issue here in  simple-system-monitor@ariel - see https://github.com/linuxmint/cinnamon-spices-applets/commit/33b76b80a3475773e9b3b568e51705fc80832abf 

I can have a look at system-monitor@ariel  if nobody else shows an interest but it may be a while before I can allocate time and it probably needs to be fixed for Cinnamon 3.8

pdcurtis:
I have found the time and hopefully fixed this issue for netusage@30yavash.com and also a few anomalies I found during testing it. 

I could only do partial testing on a LiveUSB of Fedora 27 so any feedback from users would be useful

spantaleev:
I've recently started experiencing this same problem on Archlinux.

pdcurtis:
@spantaleev  Which applet/desklet is causing the problem and what version of cinnamon are you using?

 `netusage@30yavash.com` should have been fixed and I have just had a new version merged which will work better with cinnamon 4.0

@jaszhix has also recently pushed a fix for `simple-system-monitor` for 3.8 see #327

spantaleev:
[...]

PR #443, opened 2019-07-31
simple-system-monitor@ariel - Update for Cinnamon 4.2

pdcurtis:
Added "4.2" to cinnamon-version in metadata.json to allow use with Cinnamon 4.2


brownsr:
@arielandrade

pdcurtis:
@arielandrade and @brownsr Sorry, I should have included @arielandrade but I thought he had donated it in #279 and I did the NM library fix in #360 and added version numbers etc.

PR #451, opened 2019-10-19
cpuload@kimse: Add total CPU usage switch and refresh interval

pkierski:
@ipimpat 
I've added switch for presenting total CPU usage (not separated by core). Refresh interval option has been also introduced.


ipimpat:
Thank you for your time improving this desklet.

I cloned your branch, but I cannot get it to work

I'm getting this error:

```
[cpuload@kimse]: expected expression, got keyword 'else'
[cpuload@kimse]: Error importing desklet.js from cpuload@kimse
```

When trying to add the desklet with your modifications.

```
.kimse@kimse-pc:~$ cinnamon --version
Cinnamon 3.8.9
```

pkierski:
Thank you for fast response. I've fixed (horrible! :) ) mistake.
It's weird, but automatic check missed missed this error (opening brace instead of closing one). I've added them because the same check didn't allow to use one line block without curly braces.

brownsr:
I am assuming that the positive initial response from the author means that it's fine to merge.

ipimpat:
Yes, changes are fine.

Nice work

PR #463, opened 2019-11-20
Schorschii desklet fixes

schorschii:
clock@schorschii
- use decodeURIComponent() to support paths (custom images) with spaces

battery@schorschii
- fix plug/flash icon flickering


camypaj:
wow, thanks for fixing this. I just noticed the "battery" bug, and I was about to report it, when I saw this merge request :) Kudos!

PR #515, opened 2020-04-28
googleCalendar@javahelps.com add eu.po

amaraun:
Add Basque translation (eu)


amaraun:
Add Basque translation (eu)

PR #532, opened 2020-05-31
Add Desklet system-monitor-graph@rcassani

rcassani:
I create this desklet to show graphs with the activity in various system variables (CPU, memory, and disks). Each instance of the desklet monitors one variable.


brownsr:
I suggest checking the RAM chart as it shows quite different values to the gnome system monitor

rcassani:
Thanks, I'll check that difference in reported RAM

PR #536, opened 2020-06-16
simple-system-monitor@ariel Version 1.0.1 Fix for Cinnamon 4.6 to 5.2

pdcurtis:
* Fix for Cinnamon 4.6 to 5.2 (Mint 20) by additions to cinnamon-version in metadata.json


brownsr:
Goes beyond the current cinnamon version, but that's not a problem for now.

pdcurtis:
@brownsr Yes, I am doing it deliberately on all the updates on applets I author or have undertaken to maintain as it allows the development team and beta users to use/check the applets and desklets at an early stage as well as save me and the team wasting time on a series of pull requests every 6 months when the package base is not changing. 

IF problems occur or I hear that a new Cinnamon has features which might give issues I can change back very quickly. I suppose in the ideal world there should be an option for all versions beyond a starting point but why make work.

brownsr:
Ok, quite understandable.

Regards

Simon

PR #537, opened 2020-06-16
netusage@30yavash.com Version 1.0.5 Fix for Cinnamon 4.6 to 5.2 (Mint 20)

pdcurtis:
* Fix for Cinnamon 4.6 to 5.2 (Mint 20) by additions to cinnamon-version in metadata.json


brownsr:
As before, this goes beyond the current versions of Cinnamon, but that does not cause a problem now.

Regards

Simon

PR #542, opened 2020-06-29
yfquotes@thegli: add sort option, improved connectivity

thegli:
* add setting to sort quotes list
 * update DOW 30 component symbols
 * auto-retry in case of connection problem


thegli:
@brownsr Thanks for merging my update. However, the [desklet homepage](https://cinnamon-spices.linuxmint.com/desklets/view/42) shows currently a mix of old and new files: the readme shows the new content, but the screenshot is still the old image. And the zip using the [Download button](https://cinnamon-spices.linuxmint.com/files/desklets/yfquotes@thegli.zip) contains the previous version of the desklet. Maybe I'm simply too impatient and some auto-update processes aren't complete yet?

PR #556, opened 2020-07-25
bbcwx@oak-wood.co.uk: close session after 10 seconds

Mintulix:
@tipichris 
During and after updating weather data the TCP state is ESTABLISHED, after a few seconds it changes to CLOSE_WAIT. So far ok, but it stays there "forever".
This pull request fixes this, because connection is closed after 10 seconds now.
Original fix by Gr3q (weather@mockturtl)


brownsr:
bugfix, so merging.

PR #559, opened 2020-08-13
[New Desklet] kdecdesklet@joejoetv

JoeJoeTV:
This is a Desklet, which shows information about Devices connected via KDE Connect, such as Name, Battery state and Notifications. It also allows for dismissing and responding to Notifications. It uses DBus to connect and interact with KDE Connect.

I am new to Cinnamon Spices, so I am not sure about version compatibility. What do I have to know to to make this Desklet compatible with multiple versions? How can I check if it's compatible with an older version?


JoeJoeTV:
Are these checks correct? They're saying that "imports" doesn't exist for example, but isn't that needed for creating desklets?

PR #565, opened 2020-09-26
yfquotes@thegli: Change quote percentage color based on trend

plaihonen:
@thegli 
Adds the clarity for the view by changing the color of percentage value based on the trend. (Red/Green/Inherited)
This can potentially be used with other columns as well.

### DIFF
--- a/yfquotes@thegli/files/yfquotes@thegli/desklet.js
+++ b/yfquotes@thegli/files/yfquotes@thegli/desklet.js
@@ -210,8 +210,17 @@ QuotesTable.prototype = {
     createPercentChangeLabel : function (quote) {
         return new St.Label({
             text : this.existsProperty(quote, "regularMarketChangePercent") ? (this.roundAmount(quote.regularMarketChangePercent, 2) + "%") : ABSENT,
-            style_class : "quotes-label"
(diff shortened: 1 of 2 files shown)


thegli:
@plaihonen 
Thanks for your contribution. I tried it out by myself, and I have mixed feelings. I felt that the readability of the percentage values suffers because of the colors, but that probably depends on the desklet transparency, and the desktop background. And there is already the change icon column that provides a similar quick "up/down" impression.
So, honestly speaking, I like the current layout better. But that's just my two cents...
Therefore, I will pick up your suggestion and incorporate it in my next release, but will make it configurable in the desklet's preferences. Hope that is acceptable for you.

plaihonen:
@thegli 
Indeed, you are correct. I failed to consider the desklet transparency with my change proposal.
Making it configurable could indeed solve some of the problems. Also, slight drop shadow would help.

thegli:
@plaihonen 
I integrated your change in my local branch and added a configuration setting to enable/disable the feature.
As there is currently another PR (#566) pending, it may take some days before I do release the change.
Feel free to close your PR if you're ok with my procedure.
And I leave the "slight drop shadow" improvement open for your next PR 😀

PR #566, opened 2020-09-29
Yahoo Finance Quotes: Fixes missing click support for most details

ngaro:
The yahoo finance quote of @thegli supports "click-and-surf" to yahoo finance.

But only when the user enabled "Show verbose name" was it possible to use this feature.
This PR makes it possible for all quote details, so that no matter what the user enables/disables, it will always work.

### DIFF
--- a/yfquotes@thegli/files/yfquotes@thegli/desklet.js
+++ b/yfquotes@thegli/files/yfquotes@thegli/desklet.js
@@ -134,17 +134,20 @@ QuotesTable.prototype = {
     existsProperty : function(object, property) {
       return object.hasOwnProperty(property) && object[property] !== undefined && object[property] !== null;
     },
+    addLinkIfWanted(symbol, element, addLink) {
+        if (addLink) {
+            return this.createLink(element, symbol);
+        } else {
+            return element;
+        }
+    },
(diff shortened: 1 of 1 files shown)


ngaro:
Finally, all checks passed...
Why is codacy this strict on my PR ?
When I look at the rest of the code, barely anything would end up without complaints...
Should I create a new PR with these 4 commits merged into 1 ?

thegli:
@ngaro
Thanks for your contribution. The limitation to the name column only is intentional. Some time ago, I also tried out a version with multiple clickable columns. But more often than not I encountered "click accidents" by involuntarily hitting the desklet when moving and arranging windows. That was rather distracting. Therefore the restriction to just one column, and that worked for me since then.
Therefore, I advise against making every column clickable.

But your suggestion made me think that the ticker/symbol could also be made clickable, as some people like the short quote better than the long verbose name. I will add this feature to the next release. I hope that is acceptable for you.

brownsr:
I am presuming that the new changes from the author mean that this PR is not needed. But I am leaving this open for now, in case I have not understood fully. Feel free to close it if it is not relevant anymore.

Regards

Simon

PR #567, opened 2020-10-03
Use longname if available

ngaro:
@thegli

This commit makes it use the long name instead of the short name of the quotes.

Reason:
- This doesn't change anything for well known US stocks (both the short and the long name are the same)
- Before this commit most European stocks received "ugly" short names (e.g. BFIT.AS became "BASIC-FIT".)

This commit "prettifies" those names (for BFIT.AS it would become  "Basic-Fit N.V.").
This way they don't stand and out among the other names.

### DIFF
--- a/yfquotes@thegli/files/yfquotes@thegli/desklet.js
+++ b/yfquotes@thegli/files/yfquotes@thegli/desklet.js
@@ -150,9 +150,18 @@ QuotesTable.prototype = {
             style_class : "quotes-label"
         });
     },
+    searchName : function (quote) {
+        if(this.existsProperty(quote, "longName")) {
+            return quote.longName;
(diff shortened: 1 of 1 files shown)


ngaro:
I  just noticed that you use your own repo for yfquotes-cinnamon-desklet ...
Should I sent PR's to that repo from now on ?

thegli:
@ngaro 
I incorporated your proposal into a new PR #568 (which contains parts of your other PR as well as another PR), but Codacy flagged an issue. I didn't spot a difference to the other function declarations, and will look into that later.
Please continue to create PRs on this repo.

brownsr:
Closing, author has incorporated suggestions.

Regards

Simon

Issue #544, opened 2020-07-06
[system-monitor-graph@rcassani] Text-Size for nvme-drives


schinfo:
![Bildschirmfoto vom 2020-07-06 12-59-07](https://user-images.githubusercontent.com/26834276/86586980-1d3a3480-bf89-11ea-884f-96c7f7bcd7a3.png)

 ```

@rcassani

**Issue**
Text-sice is to big for nvme-drives

rcassani:
The feature to change the text to display for variable and the size of its text will be added in a future version

rcassani:
@schinfo, The issue with the label positioning has been fixed in #575.
The Desklet also allows the use of custom labels for HDD.

rcassani:
@schinfo, can we close this issue?

PR #575, opened 2020-10-17
Fixing labels

rcassani:
This pull request does the following for the **system-monitor-graph@rcassani** Desklet
- Fix the position of the labels for the CPU, RAM and HDD graphs.
- Allow the use of custom label for the HDD graphs
- Update versioning from 1.0 to 1.1


rcassani:
This fixes #544

PR #621, opened 2021-02-23
[xkcd@rjanja] Settings for xkcd comic viewer

AndreiMiculita:
This desklet often becomes larger than the screen, as it doesn't have a maximum height and width option.
Furthermore, its readme instructs the user to just edit the metadata.json to configure the desklet. Which is... not very nice.

To fix these issues, I've added a settings window.
The settings are:

* maximum height
* maximum width
* refresh interval
* whether to keep the desklet centered (otherwise, keep it at the top left of its bounding box)

Working on this, I also migrated it to use ClutterImage instead of ClutterTexture (which is deprecated).

As of now, the main issue is getting the image to behave as expected:

- [ ]  only resize when at least one of the size limits is exceeded (so, don't resize if the image is smaller than the dedicated space)
- [ ]  set gravity to center when `keep-centered` is true, otherwise set gravity to top-left
- [ ]  add back tween animation that plays when the image is changed

There is still some unexpected behavior, I am probably misunderstanding the Clutter documentation. If anyone wants to help, do feel welcome to suggest edits.

### DIFF
--- a/xkcd@rjanja/files/xkcd@rjanja/desklet.js
+++ b/xkcd@rjanja/files/xkcd@rjanja/desklet.js
@@ -76,6 +76,8 @@ XkcdDesklet.prototype = {
     updateUI: function(xkcdId) {
         let url, filename;
 
+        this._clutterBox.set_size(this.maxWidth, this.maxHeight)
+
         if (xkcdId === null || xkcdId === undefined) {
             url = 'http://www.xkcd.com/info.0.json';
             filename = this.save_path + '/temp.json'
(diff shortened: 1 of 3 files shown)


brownsr:
Note this is still draft.  As such it can't be merged.  Do you intend this to still be in a draft state ?

Regards

Simon

AndreiMiculita:
Hey,

I left it like this on purpose, as it's not done yet. Lately, I haven't had the time to work on this. There are still a few things I need to fix, and I still intend to do that. I hope this isn't blocking any other people who might want to work on it.

brownsr:
Appears abandoned.  Feel free to reopen if work restarts.

Regards

Simon

PR #626, opened 2021-03-13
Replaces 'contains' with 'includes'.

icarter09:
Desklet was failing to install correctly. Replacing "contains" with "includes" fixes the issue and the desklet is working as expected.

@mohammad-sn


icarter09:
tagging @NikoKrause in case the author doesn't respond

PR #642, opened 2021-04-11
Adjusted hu.po files for the desklets

bossbob88:
Hi,

Some wording had to be adjusted in the Hungarian translation files.


NikoKrause:
Nice job translating the applets and desklets.

Are you aware of the translation status tables? 
They help finding untranslated strings, e.g. missed during translation or if an applet/desklet has been updated and introduces new strings, which need translation.

You find them in the README at the very bottom.

https://github.com/linuxmint/cinnamon-spices-desklets/blob/translation-status-tables/.translation-tables/tables/hu.md

bossbob88:
Many thanks Niko!

Yes, I'm aware of that. I will check them regularly if my time allows to update the translations to the newest versions.

NikoKrause:
Sometimes there's a chance that an applet or desklet is missing a string in the translation template or the translations are not applied, because the file has a bug or is not properly prepared for translation. If you notice such cases while translating, you can create an issue and ping me. I'll have a look then.

PR #651, opened 2021-06-28
Update nepalipatro.js

shres58tha:
found that the date needs +1 for fixing the date in Nepali as it currently lags a day in local Nepal time at Kathmandu, Nepal. 
I am using this desklet after fixing it manually as  done here.
Thank you. 
This  is a really nice and the most useful desklet for people of Nepal


brownsr:
ping @kanchudeep

PR #665, opened 2021-09-20
calc@ptandler: Mini-Calc Desklet

ptandler:
Very simple and minimalistic but yet powerful calculator that evaluates JavaScript expressions.
I developed it because I was missing a very quick and easy way to do some simple calculations. And JS turned out to be quite handy ;-)


ptandler:
@brownsr thanks for merging so quickly! :+1:

PR #668, opened 2021-09-27
shutdown@phpdreamer: add pt_BR.po

marceloaof:


brownsr:
Thanks for all the translations !

Simon

PR #671, opened 2021-09-27
TimeAndDate@nightflame: add pt_BR.po

marceloaof:


brownsr:
You have certainly been busy !!

SImon

marceloaof:
Yes @brownsr the mission to improve people's lives never ends.
Thank you for approving the translations.

Simon, I would like to ask you one question. Mint Update Manager already brought this merged translation to the computer. The desklet is translated and its description is also translated on the first tab of the Desklets window. However, on the second tab of the Desklets window, the same desklet-description remains to be read in english language. Is there anything I need to do to also translate the desklet description for the second tab? This question also applies to Extensions and Applets. They become translated, but their names and descriptions remain to be read in english or with old translations on the second tab. Clicking on that Refresh Cache button below the Download list had no effect.

image

PR #739, opened 2021-12-14
bbcwx@oak-wood.co.uk: Convert HTTP services to HTTPs

gwillgues:
Several of the weather API data sources in bbcwx@oak-wood.co.uk use HTTP for connections when HTTPS (TLS) connections are available. This is negative because it enhances the ability to track users who utilize this desklet. 

Changing these URLs from HTTP:// to HTTPS:// will resolve this issue.

I went through and changed the URLs and verified the back end services support HTTP, there should be no impact, it is working and connecting over HTTPS as expected on Debian unstable. 

This mitigates tracking of users (leakage of globally unique API keys in plaintext on an interval)


brownsr:
ping @tipichris

brownsr:
ping @tipichris

PR #748, opened 2021-12-27
yfquotes@thegli update

sulonetskyy:
This PR updates the yfquotes@thegli desklet.
The common list of changes:
- set symbolic icons instead of .svg
- added configurable trend change colors settings instead of hardcoded colors (NOTE: lang translations should be updated with new settings)
- fix number rounding (always rounded instead of in some cases)
- adds UAH and RUB currency symbols
- change table items view (text align to the left, numbers/dates align to the right)
- fix label/button representation (set button text align to the left instead of center - need review, maybe someone can propose a better solution !)
- refactoring and cleaning


sulonetskyy:
@thegli Could you please review? Thanks )

thegli:
@sulonetskyy Thanks for your contribution. I just started with the review, so please be patient.
With which Cinnamon versions did you test?

sulonetskyy:
@thegli I use Cinnamon 5.0.7 on Linux Mint 20.2

thegli:
@sulonetskyy I run the same distro and version. I didn't find time yet to test with older Cinnamon versions, but I went trough your list of changes, tried out your code, and here's my feedback:

- _set symbolic icons instead of .svg_
The vertical alignment issue with the image files is now gone. Neat :smiley: 

- _added configurable trend change colors settings instead of hardcoded colors (NOTE: lang translations should be updated with new settings)_
The possibility to select your own colors for the up and down trend is a cool idea. Wonder why this never crossed my mind.
**Change request**: Please also add the color for the "unchanged" case to the settings, with some gray-ish color as default.

- _fix number rounding (always rounded instead of in some cases)_
The rounding was also done before but the trailing zeros were dropped. Fixed positions make the column easier to read and it presents itself in a more orderly fashion.
I also played around with that some time ago but then decided against a fixed number of decimal places because of the "space waste". For some instruments you want a high precision (e.g. exchange rates), and for others, two or three decimals places might suffice. With the old logic you could increase the maximum number of positions but they only kick in when needed.
I don't feel comfortable to change the rounding to a fixed position mode the "hard way".
**Change request**: Make the fixed number of positions of decimal places optional by adding a setting for it. I leave it to you to decide what the default value of that option is - you know my preference :wink: 

- _adds UAH and RUB currency symbols_
:+1: 

- _change table items view (text align to the left, numbers/dates align to the right)_
The layout looks definitely more organized. On the downside, the desklet width requires more space, because of the left-right ordering, and the line might be hard(er) to read, the bigger the difference is between the shortest and the longest instrument name. But if we keep the fixed number of decimal places an optional feature (see above), then I'm fine with the "new order".

- _fix label/button representation (set button text align to the left instead of center - need review, maybe someone can propose a better solution !)_
Honestly, I don't have any better idea for now. 
**Fix request**: The link for the instrument's name got broken as now the labelText is added instead of the (unique) symbol. Please change this back.

- _refactoring and cleaning_
Cleaner code is always appreciated, thanks.

sulonetskyy:
@thegli I made changes according to the review. Could you please review it again?
List of changes:
- added configurable unchanged trend color settings
- revert the original number rounding and add the strict rounding as a parameter
- fix labelText is added instead of the (unique) symbol

About the last commit, I didn't revert the changes of "set button text-align to the left instead of center", but only fix the issue related to the "labelText is added instead of the (unique) symbol". For now, the button text is still aligned to the left. Should I revert the alignment to the center?

Thanks )

thegli:
@sulonetskyy The latest changes look all good. And it's fine by me to stick with the right alignment.

I also tested all changes with LM 19.0 and I couldn't find any problems (not that I was expecting any).
 
To make it complete, maybe you bump the version in metadata.json (e.g. 0.7.0), and add the list of changes to the release notes section in the README.md.

sulonetskyy:
@thegli, done. Thanks )

PR #753, opened 2022-01-12
yfquotes: add ru.po

sulonetskyy:
This PR updates language files of the yfquotes@thegli desklet: 
- add/remove fields to the existed languages according to the last version of the desklet
- add ru.po translation

### DIFF
--- a/yfquotes@thegli/README.md
+++ b/yfquotes@thegli/README.md
@@ -20,6 +20,10 @@ Check out the desklet configuration settings, and choose the data refresh period
 
 ## Release Notes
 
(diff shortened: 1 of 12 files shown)


sulonetskyy:
@thegli, please review )

sulonetskyy:
@thegli, with the last commit I revert my changes of the change percentage trend colors in the case when changes less 0.001% because the results are not correct. Please review )

thegli:
@sulonetskyy Looks good to me. And just in case you didn't know: PRs that only add or update translations don't need the desklet author's consent.

PR #761, opened 2022-04-17
system-monitor-graph@rcassani: German translation

schorschii:


schorschii:
German translation is included in the new PR #763

Issue #762, opened 2022-04-17
system-monitor-graph@rcassani: HiDPI display issue


schorschii:
system-monitor-graph@rcassani

```
 * Desklet version/Build date: Version 1.2 (2021-12-06 10:45:56)
 * Cinnamon version: 5.2.7
 * Distribution: Mint 20.3
 ```

**Notify author of desklet**
@rcassani 

**Issue**
Text is overlapping the graph on HiDPI displays (when UI scaling is set to 2x in Cinnamon settings).
[The desklet should take `global.ui_scale` into account](https://github.com/linuxmint/cinnamon-spices-desklets/issues/326).

![Bildschirmfoto vom 2022-04-17 18-07-56](https://user-images.githubusercontent.com/17480795/163723470-9da8b4f3-c497-461d-9709-32d6acae52b2.png)

P.S. Thanks for this desklet. I really like it :)

PR #763, opened 2022-04-17
system-monitor-graph@rcassani: add option to display swap usage

schorschii:
This adds the option to display the swap usage and a German translation.
@rcassani , what do you think?

### DIFF
--- a/system-monitor-graph@rcassani/files/system-monitor-graph@rcassani/desklet.js
+++ b/system-monitor-graph@rcassani/files/system-monitor-graph@rcassani/desklet.js
@@ -216,7 +216,7 @@ SystemMonitorGraph.prototype = {
         }
 
         // concatenate new value
-        values.push(value);
+        values.push(isNaN(value) ? 0 : value);
         values.shift();
         this.values = values;
 
--- a/system-monitor-graph@rcassani/files/system-monitor-graph@rcassani/desklet.js
+++ b/system-monitor-graph@rcassani/files/system-monitor-graph@rcassani/desklet.js
@@ -283,18 +283,27 @@ SystemMonitorGraph.prototype = {
         });
 
         // text position and content
-        this.text1.set_position(unit_size, (2.5 * unit_size) - this.text1.get_height());
+        this.text1.set_position(
+            Math.round(unit_size),
+            Math.round((2.5 * unit_size) - this.text1.get_height())
+        );
         this.text1.set_text(text1);
         this.text1.style = "font-size: " + text1_size + "px;"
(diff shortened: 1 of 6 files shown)


system-monitor-graph@rcassani/files/system-monitor-graph@rcassani/settings-schema.json
```diff
@@ -125,6 +126,13 @@
         "dependency": "type=ram",
         "tooltip": "RGB or RGBA. Alpha is ignored"
     },
+    "line-color-swap": {
+        "type": "colorchooser",
+        "default": "rgba(137,190,67,1.0)",
```
rcassani: Can you change the default color to: 
`rgba(229,165,10,1.0)`

rcassani:
@schorschii, it looks nice!
I think we can use this PR to correct the issue #762
Can you add me as collaborator to your repo to push there?

schorschii:
I corrected the issue with the last commit :)

system-monitor-graph@rcassani/files/system-monitor-graph@rcassani/desklet.js
```diff
@@ -120,7 +120,7 @@ SystemMonitorGraph.prototype = {
         }
 
         // Desklet proportions
-        let unit_size = 15 * this.scale_size;  // pixels
+        let unit_size = 15 * this.scale_size * global.ui_scale;  // pixels
```
rcassani, line 123: `unit_size` does not need modification. 
The size of the graph (in pixels) should remain the same regardless `global.ui_scale`. Only text size needs to be scaled.

system-monitor-graph@rcassani/files/system-monitor-graph@rcassani/settings-schema.json
schorschii: No problem, I changed the default color with the latest commit.

system-monitor-graph@rcassani/files/system-monitor-graph@rcassani/desklet.js
schorschii: I think it would be better to take `global.ui_scale` into account for `unit_size` because otherwise the desklet will appear unexpectedly small by default on HiDPI screens.

rcassani: Yes, It'll look small although that is the expected behaviour. The `Scale factor` in the Desklet settings indicates that `1 = 330×120 px`

PR #774, opened 2022-06-09
notes@schorscii: Add option to set custom background color

GlowingScrewdriver:
This commit provides a settings option to set a custom background color when the note style is set to "None". The default background color is transparent, so that the desklet behaves like before if a user does not want to make use of the option.

@schorschii


schorschii:
Looks good, thanks for this improvement!

schorschii (approved):

GlowingScrewdriver:
Any progress on this?

PR #776, opened 2022-06-11
system-monitor-graph@rcassani - Use asynchronous wait when taking measurements

rcassani:
It fixes as perceptible pause when the GPU readings were used. 

This PR updates the desklet to use `wait_asynch` to wait for getting the different measurements (CPU, RAM, HDD, and GPU). The code refactoring  was inspired by the [diskspace@schorschii](https://github.com/linuxmint/cinnamon-spices-desklets/tree/master/diskspace%40schorschii/files/diskspace%40schorschii) desklet.

I have tested all the measurements except for `Swap`, as I don't have access to a system with swap partition. It would be good to test this before merging.


fredcw:
Swap graph seems to be working fine.

PR #780, opened 2022-07-08
system-monitor-graph@rcassani - Added a option to display CPU temps 

20Nick:
I have added a option to display the CPU temp. This grabs the temperature from `/sys/class/hwmon/hwmon0/temp1_input`.

Preview of added feature CPU Temp

@rcassani


rcassani:
1. It seems that from kernel > 5.3, reading CPU temperatures with the suggested file is not available any more. It would be better to use the `sensors` command ([lm-sensors](https://github.com/lm-sensors/lm-sensors)) to get that information.

2. Also, there is the PR #776, in which the variable values are async read. Thus I think it's better to wait until the asynch PR is merged before adding other variables.

PR #795, opened 2022-09-07
xkcd@rjanja: Added support for localized special directories name 

Sayanel01:
@rjanja

The default savepath for images is `~/Pictures`. If your system was setup in another language, this folder may not exist, for example it's `~/Images` instead, in french.

With this PR, the default value for metadata.directory is now `"<DIRECTORY_PICTURES>/xkcd"`, which will find the right default picture folder on the system.

Possible values are the list of special directories found here https://gjs-docs.gnome.org/glib20~2.66.1/glib.userdirectory

Nb:
- Special directory is optional. `~/xkdc` or `Pictures/xkcd` will still work.
- Program looks for special name within `<...>` at the start of the path. 
- To use special name, path must start with it (`/home/<DIRECTORY_PICTURES>/xkcd` is not valid ; `<DIRECTORY_PICTURES>/xkcd` is valid)
- Invalid special name will be replaced by the default `~/Pictures`

### DIFF
--- a/xkcd@rjanja/files/xkcd@rjanja/desklet.js
+++ b/xkcd@rjanja/files/xkcd@rjanja/desklet.js
@@ -230,6 +230,22 @@ MyDesklet.prototype = {
             }));
 
             let dir_path = this.metadata["directory"];
+
+            // Replace special directory name if any, e.g. DIRECTORY_PICTURES/xkcd replaced by /home/username/Pictures/xkcd.
+            let special_dir_name = dir_path.substring(dir_path.indexOf("<") + 1, dir_path.indexOf(">"))
+            if (dir_path.indexOf("<") == 0 && special_dir_name != "") {
+                let special_dir_enumvalue = eval("GLib.UserDirectory." + special_dir_name)
+
+                // If invalid speciale dir name, get_user_special_dir falls back to desktop folder. Here we default to ~/Pictures
+                if(special_dir_enumvalue != undefined) {
+                    let special_dir_path = GLib.get_user_special_dir(special_dir_enumvalue)
+                    dir_path = dir_path.replace("<" + special_dir_name + ">", special_dir_path);
+                }
+                else {
+                    dir_path = dir_path.replace(/\<.*\>/g, "~/Pictures"); // Regex capture all between "< >"
+                }
+            }
+
             this.save_path = dir_path.replace('~', GLib.get_home_dir());
             let saveFolder = Gio.file_new_for_path(this.save_path);
             if (!saveFolder.query_exists(null)) {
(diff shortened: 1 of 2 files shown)


fredcw:
I don't see the point of saving the pictures in the first place. If people want a copy of the picture, they can just go to the xkcd website and save the image. The point of the desklet is only to save going to the xkcd website to quickly see the latest comic. I think the save picture feature should just be removed altogether. Also, if you are going to keep the feature, the problem with putting a user configurable value in metadata.json is that the value will be overwritten each time the desklet is updated because all the files in the desklet's directory are replaced with the new versions. Maybe would be better to not make the save path configurable at all and just set it to `GLib.get_user_special_dir(GLib.UserDirectory.DIRECTORY_PICTURES) + "/xkcd"`

Sayanel01:
Well it would be better to have a setting window and have user set their preference there. The #621 do add this, but it doesn't seem to have made any progress recently, so I'll add my own setting window to be merged with @AndreiMiculita latter.
This does not prevent the merge of this PR.

And regarding removal of this folder altogether... I guess the original idea was to limit data transfer (when reviewing older xkcd, which is not the main point of this desklet) but that's cheeseparing. Or just that @rjanja wanted to have all xkcds saved locally in case of global apocalypse. I guess we could remove it, or add it as an option in a setting window.

fredcw:
The main priority is to make it compatible with Mint 21. In Mint 21, I get the error `Clutter.Texture is not a constructor` and the desklet doesn't work at all. Also the tooltips need to be fixed because they don't work in Mint 20. The Desklet also uses `Cinnamon.write_soup_message_to_stream()` which no longer exists since cinnamon 5.4.10. I think we should consider @AndreiMiculita's PR as abandoned since it hasn't been updated since feb 21 and IMO, no config options should be necessary for such a simple xlet.

Sayanel01:
Well that's right, but goes way beyond this PR...

rjanja:
Fantastic, thank you @Sayanel01 for the updates 🙏 

> Or just that @rjanja wanted to have all xkcds saved locally in case of global apocalypse.

🤣 I think I may have had it downloading each image so that when Cinnamon starts there wasn't a lag time between the desklet appearing and when the image was retrieved from the web and could then be displayed. If I remember correctly, the behavior before that was a white rectangle showing on the desktop for a moment or two. Navigating forward and back was also taking some time and images were not being cached. I'm not sure what the behavior is today but I trust the decisions of this capable community when it comes to making applets like this more usable 😄 

I agree that the Mint 21 compatibility is an entirely separate issue. I'm not sure if an issue has been created here on Github to track it, yet.

rjanja (approved):
LGTM!! 🚀

Sayanel01:
Thanks :) As a first-tim contributor I need a maintainer to approve. Does that means you @rjanja, or should I tag someone else?

fredcw:
@Sayanel01 I think his approval was implied:

>...I trust the decisions of this capable community when it comes to making applets like this more usable...

It's just a matter of waiting until someone merges it.

Issue #821, opened 2022-12-19
xkcd@rjanja didn't work


christian179717:
```
 * 2022-09-07 14:30:17
 * Cinnamon 5.4.12
 * Mint 21
 * Intel Corporation HD Graphics 5500
 * Snippet or attachment of ~/.cinnamon/glass.log not present
 * Snippet or attachment of ~/.xsession-errors nothing related to that

 ```

**Notify author of desklet**
@rjanja (was not find by github)

**Issue**
the desklet didn't show anything
![grafik](https://user-images.githubusercontent.com/89655752/208460526-3710ffd6-7eb1-4be4-92f0-3f1950fc68be.png)

**Steps to reproduce**
- add the desklet to the homesreen
- nothing is shown

**Expected behaviour**
the comic viewer should show the last comic

**Other information**
In the desklet overview a attention label is shown:
![grafik](https://user-images.githubusercontent.com/89655752/208461114-c598e21a-69a8-4bc5-88eb-10969a29b03c.png)

PR #823, opened 2023-01-02
[xkcd@rjanja] Fix for mint 21

Sayanel01:
The desklet is broken on mint 21 due to newer version of CJS:
 * newest Clutter version removed `Clutter.Texture` : Replaced `Clutter.Texture` by `Clutter.Image` (for display the comic) + `GdkPixbuf` for loading image from file.
 * newest Soup version removed `Cinnamon.write_soup_message_to_stream(...)` : fixed using `stream.write_bytes()`

Also removed useless and now broken tooltip part

Also fixed warning about `Mainloop.source_remove()` that showed up in Cinnamon log due to `this._timeoutId` not being unset.
 
Nb: I couldn't find which versions are used by CJS exactly. Looks like Soup 2.4 and Clutter above 7.

Nb 2 : Clutter project is now EOL (https://blogs.gnome.org/clutter/) so eventually it will have to be replaced.

Tested under Mint 20 and 21

### DIFF
--- a/xkcd@rjanja/files/xkcd@rjanja/desklet.js
+++ b/xkcd@rjanja/files/xkcd@rjanja/desklet.js
@@ -1,7 +1,6 @@
 const Gio = imports.gi.Gio;
 const St = imports.gi.St;
 const Desklet = imports.ui.desklet;
-const Lang = imports.lang;
 const Mainloop = imports.mainloop;
 const Clutter = imports.gi.Clutter;
 const GLib = imports.gi.GLib;
@@ -14,8 +13,13 @@ const Tooltips = imports.ui.tooltips;
 const PopupMenu = imports.ui.popupMenu;
 const Cinnamon = imports.gi.Cinnamon;
 const Soup = imports.gi.Soup
-let session = new Soup.SessionAsync();
-Soup.Session.prototype.add_feature.call(session, new Soup.ProxyResolverDefault());
+let session
+if (Soup.get_major_version() == "2") {
+    session = new Soup.SessionAsync();
+    Soup.Session.prototype.add_feature.call(session, new Soup.ProxyResolverDefault());
+} else { //Soup 3
+    session = new Soup.Session();
+}
 
 const Gettext = imports.gettext;
 const UUID = "xkcd@rjanja";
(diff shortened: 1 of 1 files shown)


Sayanel01:
Note: fix for #821

mtwebster:
Should this be versioned so it still works with Cinnamon < 5.4?

Also, soup 2 is going away - spices need to be ported to use soup3. Some details here:
https://libsoup.org/libsoup-3.0/migrating-from-libsoup-2.html

If you search thru the -applets, -desklets and -extensions repos you'll  see lots of commits for migrating - you can probably work from those (it's not a huge amount of work).

We do use api 7 of clutter. Clutter isn't going away, its been moved into the muffin package.


The gjs doc site is nice - https://gjs-docs.gnome.org/

You can import this config file there to set it up for most libraries used in Cinnamon (renamed to .txt so github would allow it):
[devdocs-cinnamon.json.txt](https://github.com/linuxmint/cinnamon-spices-desklets/files/10491921/devdocs-cinnamon.json.txt)

Sayanel01:
> Should this be versioned so it still works with Cinnamon < 5.4?

I also tested with mint 19 Tara (Cinnamon 3.8.8), and it worked, so no. I didn't tested earlier version

> spices need to be ported to use soup3

I can't find a way to test with libsoup-3.0. I installed `libsoup-3.0-0` package and [tried](https://gjs.guide/guides/gjs/intro.html#in-a-gtk-application)

    imports.gi.versions.Soup = "3.0"
    const Soup = imports.gi.Soup

But that doesn't seems to be it: `Soup.get_major_version()` always return `2`. What should I do to test with soup 3?

Sayanel01:
@mtwebster

badmin-c:
I just downloaded the fixed desklet.js and replaced the old one. xkcd Desklet still does not work. Do I need something else changed?

Sayanel01:
I think you need to restart cinnamon or reboot.

If that's not enough, I'll need more info. Which version of mint and cinnamon do you have?
Can you open looking glass (win+L), under "extensions", search xkcd, right click -> reload code
Then give me the end of the log ?

badmin-c:
> I think you need to restart cinnamon or reboot.

Oh my... You're right. It works after reboot. Shame on me ^^

Thank you for your time!

PR #829, opened 2023-02-12
mkbhru: googleCalendar@javahelps.com added hi.po

mkbhru:
I've added a Hindi translation for this desklet.
I am from India and I love this desklet.
Yearning to see this translation in the desklet as soon as possible.


mkbhru:
@slgobinath

PR #816, opened 2022-12-11
Simple system monitor: version 1.1.0

kancolle386:
I love this desklet. So, I maintained for more usefull.

- Add README.md to organize the User's Tips.
- Add function.(GPU Temperature)
- Adjust layout. (Labels: right to left, Values: left to right)
- Adjust value format to avoid wobbling. ("CPU" "Memory" fixed decimal point to 2 digits)
- Adjust value format to avoid wobbling. ("Download" "Upload" fixed decimal point to 1 digits in unit is "MB")

PR #832, opened 2023-03-26
simple-system-monitor@ariel: Add configurable options for user customization

rcalixte:
After #816, making the settings configurable through the user interface rather than requiring manual file modifications for Cinnamon 5.8+.

Also breaking out a dedicated CHANGELOG and other general cleanup.

cc: @kancolle386 @brownsr @arielandrade


rcalixte:
Given the potential for disruption, I've opted to rename the `5.6` folder to `5.8`. If **[deb3926](https://github.com/linuxmint/cinnamon/commit/deb3926b44ee4268d1051dc668be32762991ba43)** gets a release and then picked up by enough distributions, we can always reset this.

If anyone wants to test/run this locally on Cinnamon 5.6 until then, rename the desklet folder to `5.6` and applying the referenced commit locally (or build from source since it has been merged upstream).

brownsr:
No response from original author.  Merging as it seems a beneficial upgrade from a future maintenance point of view.

Simon

PR #846, opened 2023-05-13
show-remote-ip-desklet@nejdetckenobi: Add soup 3 compatibility, plus a few improvements.

mtwebster:
- Added proper settings
- Allow showing ip4, ip6 or both
- Fix urls so correct ip address type is reliably returned.

@nejdetckenobi


nejdetckenobi (approved):

nejdetckenobi:
Thank you @mtwebster . I didn't have time to add new features. It works just fine.

PR #855, opened 2023-06-20
[notes@scorchii] Avoid Array.toString() error messages

claudiux:
@schorschii 

This PR avoid Array.toString() error messages appearing in `~/.xsession-errors`, like:
```
(cinnamon:2062): Gjs-WARNING **: 23:25:36.963: Some code called array.toString() on a Uint8Array instance. Previously this would have interpreted the bytes of the array as a string, but that is nonstandard. In the future this will return the bytes as comma-separated digits. For the time being, the old behavior has been preserved, but please fix your code anyway to explicitly call ByteArray.toString(array).
(Note that array.toString() may have been called implicitly.)
```

Please, see https://github.com/orgs/linuxmint/discussions/48

### DIFF
--- a/notes@schorschii/files/notes@schorschii/desklet.js
+++ b/notes@schorschii/files/notes@schorschii/desklet.js
@@ -19,270 +19,306 @@ const DESKLET_ROOT = imports.ui.deskletManager.deskletMeta[UUID].path;
 
 // translation support
 function _(str) {
-	return Gettext.dgettext(UUID, str);
+  return Gettext.dgettext(UUID, str);
 }
 
+const to_string = function(data) {
+  return ""+stringFromUTF8Array(data);
+}
+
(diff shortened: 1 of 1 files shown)


schorschii:
Thanks bringing this up, but why such a complicated fix with an own `stringFromUTF8Array()` implementation and not simply use `ByteArray.toString(contents)` as suggested by the warning message?

Furthermore, I'd like to keep tabs for indentation instead of spaces.

My proposal: https://github.com/schorschii/cinnamon-spices-desklets/commit/5bf025d5cad703afaff389ecf5437bf8c9ba71f7

schorschii:
Yes, I created the PR https://github.com/linuxmint/cinnamon-spices-desklets/pull/856 from my branch. Thanks!

PR #860, opened 2023-07-07
dual-datetime: Initial release

rcalixte:


claudiux:
Works well. All that's missing is a transparency setting.

claudiux:
@rcalixte Ah there is a bug:
Capture d’écran du 2023-07-23 12-41-53

rcalixte:
> @rcalixte Works well. All that's missing is a transparency setting.

Transparency can be set in the color options? I provided similar instructions [here](https://github.com/linuxmint/cinnamon-spices-applets/issues/4809#issuecomment-1627834373). There are font color options and the decorations option and both use the standard color chooser. Is there something I'm missing?

> @rcalixte Ah there is a bug

This is actually intentional. There's a setting in the options for the width but I intentionally didn't want the desklet's appearance to be impacted if the text was too big or too wide. (This is similar to the other `Pango` text options from last week.) There's a width setting that allows extending beyond the text as well so that it can be positioned however the user wants next to other desklets. This also works with the 'Horizontal' setting. I admit it's a little bit unorthodox but when I took a look at other time-based desklets, it was a constant feature I saw people asking for in the comments.

Thanks for testing! Please feel free to share any other findings!

claudiux:
I was talking about background transparency.

> This is actually intentional.

Ah, ok.

Issue #840, opened 2023-04-20
yfquotes@thegli  Error: Yahoo Finance service not available!


David-DeZ:
Version 0.8.2 (2023-03-31 12:27:54)
Linux Mint 21.1 Cinnamon

Noticed  this morning (20/4/23) that it is blank not listing my stocks but with, in red, "Error: Yahoo Finance service not available!". Yahoo Finance via the web (https://finance.yahoo.com/portfolios) does work, however they have introduced a beta test. Also yesterday updated Linux Kernal to 5.15.0-70 from 5.15.0-69.  Firefox & Chromium were also updated.
**Notify author of desklet**

David-DeZ:
As quickly as the problem came up, it has now disappeared - however I won't close it yet in case it reappears on a daily basis -  perhaps it is based on when the market is open or closed - the market is now open and it's fine, this morning (GMT+3) when it was closed the problem existed. 

I will close this issue after a few days if all continues well.

thegli:
The problem we encountered a few weeks ago unfortunately reappeared. AFAIK it has nothing to do with the market opening period, but with a technical change Yahoo did on their side. So far I'm still optimistic that they again will resolve the issue shortly.

BTW, the problem is _not_ related to Linux Mint or Cinnamon.

thegli:
Created PR #843 which allows to switch between API versions

David-DeZ:
Thanks Thomas!

David-DeZ:
Thanks so much, Thomas!

Any idea when the change will be distributed via Linux Mint? 

Much appreciated!

thegli:
That I cannot say. The PR is now in the hands of the repo maintainers, like @brownsr

In the meantime, you can patch the desklet code yourself. The small procedure (a single command + restart) is documented in the [desklet's comment section](https://cinnamon-spices.linuxmint.com/desklets/view/42).

teto90:
Great Job as ever!

Thanks, I was experiencing the same problem!

David-DeZ:
Sorry to report that the problem has returned - May 24 05:20 GMT
I redid the terminal command and restarted Cinnamon but this time it didn't work.

thegli:
PR #853 should take care of this (for V7 only - no fix for V6)

David-DeZ:
Thanks!  Any "guesstimate" when it might be available?

thegli:
Your guess would be as good as mine :smile:
[...]

PR #864, opened 2023-07-19
yfquotes@thegli: add setting to configure HTTP User-Agent request header

thegli:
This new, optional setting may help in high-load situations.

### DIFF
--- a/yfquotes@thegli/files/yfquotes@thegli/po/de.po
+++ b/yfquotes@thegli/files/yfquotes@thegli/po/de.po
@@ -1,5 +1,5 @@
 # German translation for Yahoo Finance Quotes Desklet
-# Copyright (C) 2021 Thomas Egli
+# Copyright (C) 2023 Thomas Egli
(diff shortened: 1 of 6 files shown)


yfquotes@thegli/files/yfquotes@thegli/po/de.po
```diff
@@ -1,13 +1,13 @@
 # German translation for Yahoo Finance Quotes Desklet
-# Copyright (C) 2021 Thomas Egli
+# Copyright (C) 2023 Thomas Egli
```
rcalixte, line 2: Just a comment, you normally want this to be the year that the code originated, not when it was updated. You can use a range too if you'd like but that's just extra work.

rcalixte (approved):

yfquotes@thegli/files/yfquotes@thegli/po/de.po
thegli: @rcalixte Thanks for the hint. And highly appreciate the super-fast approval :100:

rcalixte:
May resolve #840

PR #909, opened 2023-08-15
InternetTimeDesklet@stefan: Fix timezone for accurate Beat Time calculation

christopher-conley:
Change the hour/minute/second variable assignments to explicitly be in UTC format. Also, add one hour to the "getUTCHours()" call, as "Biel Mean Time" (BMT), which Beat Time is based upon, is defined by Swatch as UTC+1.

Previously this was calling Date's getHours/getMinutes/getSeconds function family, which returns _LOCAL_ time. This had the effect of the Beat time returned by this desklet **always** being incorrect unless your computer's time is set to the UTC+1 timezone.

This commit fixes that and ensures that the correct universal Beat Time will be displayed regardless of the user's timezone.

Original author of this desklet is @stefan12O

### DIFF
--- a/InternetTimeDesklet@stefan/files/InternetTimeDesklet@stefan/desklet.js
+++ b/InternetTimeDesklet@stefan/files/InternetTimeDesklet@stefan/desklet.js
@@ -74,9 +74,9 @@ NetBeatDesklet.prototype = {
 
 	refresh: function() {
 		let d = new Date();
-		let h = d.getHours();
-		let m = d.getMinutes();
-		let s = d.getSeconds();
+		let h = d.getUTCHours() + 1;
+		let m = d.getUTCMinutes();
+		let s = d.getUTCSeconds();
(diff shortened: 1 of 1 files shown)


stefan12O:
I was never aware of this and it showed always correct, as I am in the Berlin Timezone.
Your changes are plausible and I appreciate.
Thank you very much.

As you are the one who realized, do you want to do the same change to the corresponding Applet as well ?

stefan12O (approved):

rcalixte (approved):

christopher-conley:
> I was never aware of this and it showed always correct, as I am in the Berlin Timezone.
> Your changes are plausible and I appreciate.
> Thank you very much.
> 
> As you are the one who realized, do you want to do the same change to the corresponding Applet as well ?

I wasn't aware there was an applet as well, but I can make a PR for that tomorrow, sure. I'll update it to include the calculation change from a previous commit in this desklet as well.

Issue #932, opened 2023-10-21
diskspace@schorschii - desklet is causing stutter


H34dl3ss:
### Desklet version/Build date

Version 1.13 (2023-08-08 04:35:12)

### Cinnamon version

Cinnamon 5.6.8

### Distribution

Linux Mint 21.1 Vera

### Graphics hardware *and* driver used

AMD and Nvidia

### Desklet name and maintainer

diskspace@schorschii @schorschii

### What happened?

I noticed stutter when playing games in windowed mode.
When going fullscreen, the stutter is gone.

I searched for reasons and found a post in the Mint forum that stated desklets could cause the issue.
I can confirm this. An user account without desklets is not suffering of stutter.

Now I went down the rabbit hole and removed all desklets and applets and added one by one until the stutter was back.
This proved to be the diskspace applet. I have multiple drives/partitions monitored and it seems the more desklets are running, the more stutter is happening.

It is very annoying as it affects video playback also. It can be observed even with mangohud glxgears.
I'm glad to support further data if necessary.

schorschii:
Yes, I can reproduce that. It seems to get worse the more desklets are active and I could swear this didn't happen in Mint 20. I tracked it down and the problematic part seems to be the initialization of the Gio subprocess call: `subprocess.init(null);`

An alternative could be to read the information from a file since Gio's `load_contents_async()` function does not produce stuttering. So many information are available in `/sys`, `/proc` and `/run` but unfortunately, I'm unable to find file system free/total space information there.

I don't have a quick solution for this but I'll keep thinking about it.

schorschii:
OK, I think I found a solution. First, the usage of the async subprocess seems wrong. Instead of `subprocess.wait_async()`, I now use `subprocess.communicate_utf8_async()` and `communicate_utf8_finish()` as suggested by the [Gio docs](https://docs.gtk.org/gio/method.Subprocess.communicate_utf8_async.html).

Then, I found out that [Gio provides a function to get file system information](https://docs.gtk.org/gio/vfunc.File.query_filesystem_info_async.html) by itself. No need to execute a `df` subprocess anymore! A much more elegant solution.

[...]

Issue #913, opened 2023-09-01
Small but noticeable freeze in system-monitor-graph@rcassani


rcassani:
* Desklet: ` system-monitor-graph@rcassani`
 * Desklet version: `1.5`
 * Cinnamon version: `5.8.4`
 * Distribution: `Arch`

Last year I (yup, I am the creator of the desklet.) made modifications for in code of the `system-monitor-graph@rcassani`  to remove the synchronous waits (#776). However, it seems there are still calls that cause the desklet to freeze the desktop. This is noticeable when watching a video. 

![image](https://github.com/linuxmint/cinnamon-spices-desklets/assets/8238803/8596a564-02e0-435d-93ac-e91849bbc2ee)

Thus I am opening this issue to ask for support / help to solve (if possible) this behaviour, or to point towards the problem behind it.

Thanks,

fredcw:
I can't test this because I don't have an Nvidia GPU but I don't see anything in your code that would cause a delay. Even if your desklet causes cinnamon to freeze for a short time, this should not cause video to freeze in another app I don't think. I can only guess that the gpu query itself is causing the gpu to freeze.

rcalixte:
I took a look at cleaning up warnings where I could for a few applets/desklets. The challenge here is the [list of unsafe items](https://github.com/linuxmint/cinnamon/blob/master/files/usr/share/cinnamon/cinnamon-settings/bin/ExtensionCore.py#L32) that are core for this desklet's functionality, mainly GTop. I haven't found a suitable alternative for libgtop. For all of the other items in the list, there are suitable workarounds.

rcassani:
@fredcw, it seems to be related to the graph drawing. A way to replicated it is to open around 15 graphs (all CPU usage), the short-time freezing is noticeable. Although, it is not realistic to expect that many graphs.

@rcalixte, thanks for tackling the warning. I tested the #914, the behaviour seems to be the same.

Since the freezing is very short when there are few graphs, and the warning is not there anymore. This issue could be closed once the PR is merged.

PR #938, opened 2023-10-26
system-monitor-graph@rcassani: Fix Stuttering

schorschii:
Hi @rcassani , we noticed that `subprocess.init(null)` causes microstutters on our desktops, see #932. This PR fixes the issue like in my diskspace desklet (#937).

With this change, the values are read directly from files using Gio's `load_contents_async` instead of spawning `head` or `cat` subprocesses to read it.

For scenarios where this is not possible (NVIDIA GPU monitoring), I switched `subprocess.wait_async` to `load_contents_async` and `load_contents_finish` which seems to be the correct way to read subprocess output async, as described [here in the Gio docs](https://docs.gtk.org/gio/method.Subprocess.communicate_utf8_async.html).

### DIFF
--- a/system-monitor-graph@rcassani/files/system-monitor-graph@rcassani/desklet.js
+++ b/system-monitor-graph@rcassani/files/system-monitor-graph@rcassani/desklet.js
@@ -366,81 +366,80 @@ SystemMonitorGraph.prototype = {
         Mainloop.source_remove(this.timeout);
     },
 
+    cpu_file: Gio.file_new_for_path('/proc/stat'),
     get_cpu_use: function() {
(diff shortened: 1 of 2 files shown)


rcalixte:
@rcassani This seems like a noteworthy bug fix so it would be nice to have it merged sooner than later if possible. If you would like some time to review, that's fine too.

rcassani:
Hi @schorschii, nice work!
It was kind of reported in #913, which is properly solved with this PR.
@rcalixte, ready to be merged

rcalixte (approved):

rcalixte:
Thanks to both of you!

PR #964, opened 2023-12-04
diskspace@schorschii: added cpu utilization, added update widget interval, added text-view options, improved testSub generation logic

vstrizhenok:
It's nice to have cpu utilization desklet type


schorschii:
Thanks @vstrizhenok , I appreciate your effort, but the separate CPU desklet now already exists and we should focus on fixing the crash you encounter with it. Also, when searching for desklets, users wouldn't expect that a diskspace desklet can monitor the CPU.

Can you please provide more information of the CPU desklet crash in a new issue?

Nevertheless, making the update interval customizable is a nice improvement. If you remove the other changes, we can merge this.

vstrizhenok:
@schorschii 
> Also, when searching for desklets, users wouldn't expect that a diskspace desklet can monitor the CPU.
Exactly! That is my story ; ) I was surprised when I found `Disk space` desklet and discovered the ability to monitor RAM and swap but in the end I liked it ; ) 
The approach with separate desklets which monitor hardware parameters allows us to keep desklet simple but does not allow us to build one view with some aggregated information contains both disk, memory, cpu, video ram, top 5 huge processes etc. if we keep the code and view separately, all this can be configured by the user himself.
let's ignore the name of desklet or rename it if you wish.
Anyway your desklet is nice, thanks a lot  for your work!

schorschii:
Thanks for your feedback. I still have concerns about adding CPU monitoring to the diskspace desklet. You're right that RAM and swap strictly speaking are wrong here too, but RAM and swap space is thematically more close to the desklet's name than monitoring CPU.

The main thing is: what should we do with the existing CPU desklet? I don't want to maintain two desklets with (partly) same code/functionality.

vstrizhenok:
Got it, so i have to close the pull request

PR #973, opened 2023-12-06
Add Russian translation to diskspace@schorschii

blogdron:


rcalixte (approved):

blogdron:
Little question. Someone `pot` files have commentary with authors name, email, year, but if im run `./cinnamon-spices-makepot UUID`  for update `pot` file in actual (if I'm not sure `pot` actual) script rewrite it commentary... Its okey? 

I try to take the `pot` template for translation as is, if I see that there is no need to regenerate it, but sometimes I run an update. The README (which I took a while to read... :) describes the sequence in which you should always update the `pot` file. But I see many translations with very different dates and differences.

Do I understand correctly that you should always run `cinnamon-spices-makepot` to update the translation template, or should you try to do this only when absolutely necessary?

rcalixte:
There are ways to have the script generate a custom header but even I haven't done that on the Spices that I've authored. There isn't a strict rule, but for myself, if I update a .pot file on another author's xlet, I try to leave it with as much of the original context as possible. Having the updated strings for translation is the priority.



> But I see many translations with very different dates and differences.

Some of these haven't been modified since they were first added to the repository or might be abandoned by the authors.



> Do I understand correctly that you should always run `cinnamon-spices-makepot` to update the translation template, or should you try to do this only when absolutely necessary?

If there's only header changes, I'll skip adding the .pot file entirely. If there are new strings, then I'll do a merge of sorts. That's just my preference to try to preserve what the author intended.

blogdron:
@rcalixte Thank you for help, now it becomes clearer to me. I will try to transfer original additional information from the original authors (almost always this is just a comment in the title) and/or try to use the current pot file if it contains everything needed for translation, regardless of the date of its creation. I didn’t immediately notice all these nuances =)

rcalixte:
> I didn’t immediately notice all these nuances =)

It takes time! Thanks for all your work on the translations!

PR #1040, opened 2024-01-23
NumericBinaryClock@ozymandiaslone - yet another clock - a text-based binary clock :) .

ozymandiaslone:


rcalixte (approved):

ozymandiaslone:
> @ozymandiaslone If you want, you can look at the README on the main repository for instructions on how to create the template that translators can use to translate strings. We can add this for you as well if needed.

Gotcha, thanks so much! I'll look into the translations - I naively assumed I didn't need them because my desklet only displays binary numbers, but I'm realizing now they probably apply to the download page and such too 👍

rcalixte:
> > @ozymandiaslone If you want, you can look at the README on the main repository for instructions on how to create the template that translators can use to translate strings. We can add this for you as well if needed.
> 
> Gotcha, thanks so much! I'll look into the translations - I naively assumed I didn't need them because my desklet only displays binary numbers, but I'm realizing now they probably apply to the download page and such too 👍

Neither of those actually. In this case, it would be the strings in the metadata. The name and description would be translatable for easier discovery in other languages.

PR #1047, opened 2024-02-05
Rewrite cinnamon-spices-makepot:

rcalixte:
New features:
  * Automatic updating of translations when the template is updated
  * Updated header format
  * Add translation support for Glade files and shell scripts


fredcw:
@rcalixte Wow, that's some quick work.

PR #1087, opened 2024-03-13
battery@schorschii: remaining time

### LINKED ISSUE #1086: Desklet Feature Request

battery@schorschii @schorschii

### What would you like to see?

It would be great to have an option to display a battery percentage and/or time remaining to charge/discharge. This could be accomplished by changing the "Show battery in percent" option to a new "Battery display" dropdown (while migrating the existing user preference so that the upgrade is transparent). I was able to implement a similar duration option recently for the batterymonitor@pdcurtis applet and I think it would be a good addition here as well. There is also now an example of settings migration in the bbcwx@oak-wood.co.uk desklet as that was done recently too.

If the idea is amenable, I have no problem taking the time to implement it or you can do it at your own leisure. Thanks again for an awesome desklet!

schorschii:
Proposal for displaying the remaining time.
closes #1086


rcalixte:
A few local samples so far:








There was a brief point where I first plugged in this laptop and it read "(0:00)" before updating. I think this related to how the conditionals determine the final values, causing the wrong UPower property to be read in briefly initially. We could refactor this but it works for now outside of that brief period of a few seconds. Let me know if you would like to update the translation files as part of this too.

rcalixte:
I wonder if a bold style might be better for legibility?

schorschii:
Nice, thanks. Yes, a font-weight of 400 is better. I already unconsciously tested it with 400 since the light variation of DejaVu Sans is not installed per default in fresh Mint installations and so it automatically fell back to 400.

And yes, we can update the translations with this too.

rcalixte (approved):

rcalixte:
It was great working with you on this! Thanks again for the fantastic desklet!

Issue #1205, opened 2024-06-22
cpuload@kimse: Cpu load desklet crashes under all conditions


Alastair-L:
### Desklet version/Build date

0.7

### Cinnamon version

6.0.4

### Distribution

Mint 21.3

### Graphics hardware *and* driver used

NVIDIA GeForce RTX 2060 SUPER and nvidia-driver-535

### Desklet name and maintainer

cpuload@kimse @schorschii

### What happened?

Installed desklet, and when I go to activate it cinnamon crashes.

~/.xsession-errors contains the following line:
```
csd-power:1569): CRITICAL **: 20:45:29.169: abs_to_percentage: assertion 'max > min' failed
```

I believe that the error is that somehow cpus_utilization has NaN values in it. I'm not sure why or how though. CPU details:
- AMD Ryzen 7 3700X (16) @ 3.600GHz

Not sure if this is helpful, but here is the contents of /proc/stat (I wasn't able to get a sample from when the NaN values were being displayed though sorry)
```
alastair@alastair-MS-7C37:~/.local/share/cinnamon/desklets/hello-world@alastair$ cat /proc/stat
cpu  82780 65 54694 9320910 4641 0 1518 0 0 0
cpu0 4447 61 2626 584374 290 0 823 0 0 0
cpu1 3319 0 1962 586354 243 0 96 0 0 0
cpu2 4702 0 2436 584509 257 0 48 0 0 0
cpu3 4760 0 2443 584428 247 0 6 0 0 0
cpu4 6316 0 3574 581549 244 0 4 0 0 0
cpu5 6156 1 3528 581830 172 0 0 0 0 0
cpu6 5987 0 3538 581769 375 0 0 0 0 0
cpu7 6178 0 3337 582087 283 0 1 0 0 0
cpu8 4583 0 2404 584547 445 0 12 0 0 0
cpu9 2740 0 10481 574215 145 0 7 0 0 0
cpu10 4852 0 2490 584350 233 0 5 0 0 0
cpu11 4846 1 2731 584076 294 0 53 0 0 0
cpu12 6060 0 3193 582347 264 0 0 0 0 0
cpu13 6295 0 3289 581960 176 0 1 0 0 0
cpu14 5904 0 3371 581415 495 0 142 0 0 0
cpu15 5629 0 3285 581093 470 0 315 0 0 0
[...]

PR #1206, opened 2024-06-22
cpuload@kimse: Display full bar for NaN cpu_utilization

Alastair-L:
@schorschii

Reduces the impact of this bug: https://github.com/linuxmint/cinnamon-spices-desklets/issues/1205

All ok:
image
All NaN:
image
Some NaN:
image


schorschii:
I don't know if this really improves anything. We should find and fix the root cause instead. Please provide the output of /proc/stat when the NaN% occurs on some cores, so we can investigate the problem.

Alastair-L:
Thanks for your comment and for getting back to me so quickly. I've made another comment on the linked issue with what I believe to be the root cause (including the relevant line of /proc/stat and will close this PR in favor of fixing the root cause.

PR #1212, opened 2024-06-28
soundBox@scollins: Fix missing album artworks in Soundbox desklet using ibclutter v.1.12 or higher

kawashiro:
@collinss

### DIFF
--- a/soundBox@scollins/files/soundBox@scollins/soundbox.js
+++ b/soundBox@scollins/files/soundBox@scollins/soundbox.js
@@ -10,6 +10,9 @@ const Gtk = imports.gi.Gtk;
 const Cvc = imports.gi.Cvc;
 const St = imports.gi.St;
 
+const Cogl = imports.gi.Cogl;
+const GdkPixbuf = imports.gi.GdkPixbuf;
+
 const Interfaces = imports.misc.interfaces;
 const Params = imports.misc.params;
 const Util = imports.misc.util;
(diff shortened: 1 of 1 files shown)


rcalixte (approved):

fredcw:
@kawashiro The file you've edited here is only used in cinnamon versions earlier than 2.5 (very old). Since cinnamon 2.8 (Mint 17.3) the only version of the desklet that is used is in the folder `/soundBox@scollins/files/soundBox@scollins/2.8/`

kawashiro:
@fredcw Actually, all files in 2.8 directory are symlinks to a root ones so I've actually changed a 2.8 version as well. I've tried to check it on the latest Linux Mint, whooh, album arts appeared but it was broken by another reason. I've tried to fix it as well here (https://github.com/linuxmint/cinnamon-spices-desklets/pull/1218).

Regarding another open issue, I'll try to look at it later.

JS is not my primary language so I'm not sure if I could maintain this project by myself but if nobody else wants... :) I really love this desklet and do not want to have it broken

rcalixte:
> Actually, all files in 2.8 directory are symlinks to a root ones so I've actually changed a 2.8 version as well.

Be careful. This isn't true with the desklet.js files. It might be best to consolidate the folders a bit. Keep the 2.8 contents but minimize the symlinks (maybe just the css, js, and json files and symlink the rest). The remaining files should be moved from the root to a 5.4 directory. If you'd like some help on the open PR or would prefer doing it in another PR, we can do that.

fredcw:
I'd suggest removing all the symlinks. They're not preserved when the spice is installed anyway, the files are duplicated so there's no space saving and it's easier when fixing new bugs to just deal with the latest version and not worry about testing older versions when you can just presume they worked OK.

kawashiro:
Makes sense, I will do it in terms of https://github.com/linuxmint/cinnamon-spices-desklets/pull/1218 as soon as possible

PR #1214, opened 2024-06-30
yarr 1.2.2 Enhancements

jtoberling:
Yarr: ## [1.2.2] - 2024-06-28
- Enhanced logging of AI errors
- Remove tooltips from 1st buttons to improve efficiency


rcalixte:
@jtoberling For some reason, the function `formatTextWrap` appears to be breaking how strings are detected for translation. By moving it later in the class definition, there is at least a workaround. If you're okay with the finished result, we can go ahead and merge. If you'd like to push more changes or revert it, we can do that too.

jtoberling:
@rcalixte Thanks for checking! :)
I'm ok with the changes - also tested and works ok. 
Please go and merge.

rcalixte (approved):

rcalixte:
Regarding the issue, it looks like it's something in the backticks in the strings. Both instances that are in the return statement of that function introduce the same issue. I've used this notation myself and haven't seen it break string detection for translation like this so it's rather curious.

Issue #781, opened 2022-08-10
[soundBox@scollins] Volume slider has stopped working


fredcw:
Clicking with mouse on volume slider no longer does anything since cinnamon 5.4.x

Clutter.grab_pointer() not available in cinnamon 5.4.x

@collinss

collinss:
I haven't used this desklet in years. I've been meaning to remove myself as the maintainer of it for a while, but never got around to it. As with the sticky note applet, you're welcome to make whatever changes you'd like to this desklet.

PR #1218, opened 2024-07-03
soundBox@scollins: Fix sliders for the latest Cinnamon

kawashiro:
To be fixed: https://github.com/linuxmint/cinnamon-spices-desklets/issues/781


rcalixte:
@kawashiro Have you tested this on older versions of Cinnamon? We can always create a 5.4 folder and copy new code there. It would then only apply to Cinnamon 5.4+ which I believe is the breaking point here.

kawashiro:
@rcalixte Could you please suggest me a distro with such an old Cinnamon provided? Even the last Debian has 5.6.8 installed. Hope, it would not be a Nix OS or Gentoo :)

Just want to check if it really backward compatible, because regarding the docs, this API was introduced a long time ago

rcalixte:
> @rcalixte Could you please suggest me a distro with such an old Cinnamon provided? Even the last Debian has 5.6.8 installed. Hope, it would not be a Nix OS or Gentoo :)

Linux Mint 20.3 is still supported and in heavy usage. That has Cinnamon 5.2. If it works there, that should be okay.



> Just want to check if it really backward compatible, because regarding the docs, this API was introduced a long time ago

We can always merge it now and fix it later if there are issues.

kawashiro:
Yep, I really broke backward compatibility. It really does not work on 5.2. Give me some time to come up with a workaround of that

rcalixte:
> Yep, I really broke backward compatibility. It really does not work on 5.2. Give me some time to come up with a workaround of that

That already exists. We can create a 5.4 directory and put the new contents there.

kawashiro:
Make sense, will return with a fix, I hope, tomorrow

rcalixte:
> Make sense, will return with a fix, I hope, tomorrow

Shout if you need a hand and we can collaborate!

rcalixte:
@kawashiro This is phenomenal! I pushed a change to restore the 2.8 folder without symlinks (just in case) and then updated the translation files. This folder likely won't need to be touched going forward. I also renamed the 5.2 directory to 4.6 as that is the earliest version that it should support. (I should've been clearer here.) And the same for 5.6 to 5.4 as that is the boundary for the breaking changes.

Are you interested in taking over as the maintainer for this desklet? This means you would be notified for issues (but not alone).

rcalixte (approved):

kawashiro:
@rcalixte thank you for your support

> Are you interested in taking over as the maintainer for this desklet?
Let's have a try

PR #1225, opened 2024-07-11
xkcd@rjanja: Random comic selection feature

AaronDabney:
I think this would be neat 😁
@rjanja


AaronDabney:
Discovered a significant issue.  `this.curXkcd.num` does not represent the id of the websites current comic. Whoops 😁. Will republish the PR when fixed.

PR #1271, opened 2024-09-01
Add Fish completions for scripts

EmilyGraceSeville7cf:
✅ Will be available from Fish thanks to [this](https://github.com/fish-shell/fish-shell/pull/10707) PR when accepted.

# Screenshot

image

Features

For test-spice and validate-spice the following features are supported:

- Tab-completion for options

# Prerequisites

- [Fish shell](https://fishshell.com/)

### DIFF
--- a/completions/test-spice.fish
+++ b/completions/test-spice.fish
@@ -1,0 +1,4 @@
+complete -c test-spice -l help -d 'Show help'
+
+complete -c test-spice -s r -l remove -d 'Remove all test spices'
+complete -c test-spice -s s -l skip -d 'Skip validation with `validate-spice`'
--- a/completions/validate-spice.fish
+++ b/completions/validate-spice.fish
@@ -1,0 +1,1 @@
+complete -c validate-spice -l all -d 'Validate all spices'


rcalixte:
@EmilyGraceSeville7cf Both of these scripts already feature usage information when run without parameters. Again, please familiarize yourself with some of the existing tooling and please be patient for feedback. The Discussions are probably the best starting point for some of these ideas instead of Pull Requests.

EmilyGraceSeville7cf:
> @EmilyGraceSeville7cf Both of these scripts already feature usage information when run without parameters.

Yes, indeed, I've checked it before writing these simple completions. :) I'm personally Fish shell user and use Tab-completion all the time, that's why I've written these completions.

rcalixte:
> use Tab-completion all the time

The most common parameter(s) to these scripts are the UUIDs which are the subdirectories. For most modern shells, those subdirectories should already autocomplete. For the switches, there are not that many that it merits having upstream autocompletion files for shells. (You are more than fine to use this locally. No one but you can decide that.) This also ignores that the majority of distributions default to Bash still, limiting the overall usefulness of this for the community.

EmilyGraceSeville7cf:
Completions are moved to Fish shell project.

PR #1280, opened 2024-09-13
yfquotes@thegli: restore auto-refresh on import

thegli:
- cache quotes responses for each desklet instance separately
- restore automatic quotes data refresh on settings import


rcalixte (approved):

thegli:
once more, much obliged for the express approval :100:

PR #1282, opened 2024-09-16
minimalisticClock@acaimingus: Version 1.0

acaimingus:
New minimalistic clock desklet created by me.


minimalisticClock@acaimingus/files/minimalisticClock@acaimingus/desklet.js
```diff
@@ -0,0 +1,318 @@
...
+    this.installFont();
+    this.setupUI();
+  },
+
+  // Method to manually trigger font installation
+  manualFontInstall: function () {
+    let homeDir = GLib.get_home_dir();
+    let markerFilePath =
+      homeDir +
+      "/.local/share/cinnamon/desklets/minimalisticClock@acaimingus/.font_install_attempted";
```
rcalixte, line 60: You can use the GLib library for this path as well.

minimalisticClock@acaimingus/files/minimalisticClock@acaimingus/desklet.js
```diff
@@ -0,0 +1,318 @@
...
+
+    // Create marker
+    try {
+      GLib.file_set_contents(markerFilePath, ".");
+    } catch (e) {
+      global.logError("MinimalisticClock: Error writing marker file: " + e);
+    }
+
+    // Where the font should already be if installed.
+    let targetFontPath = "/usr/share/fonts/SUSE.ttf";
```
rcalixte, line 107: Don't do this. Use the XDG fonts path.

minimalisticClock@acaimingus/files/minimalisticClock@acaimingus/desklet.js
```diff
@@ -0,0 +1,318 @@
...
+      if (success) {
+        global.log("MinimalisticClock: Font installed successfully.");
+
+        // Reload Cinnamon
+        let restartCommand = ["cinnamon", "--replace"];
+        let restartProc = new Gio.Subprocess({
+          argv: restartCommand,
+          flags:
+            Gio.SubprocessFlags.STDOUT_PIPE | Gio.SubprocessFlags.STDERR_PIPE,
+        });
```
rcalixte, line 165: This isn't how you reload a font cache and could be more disruptive than doing it properly.

minimalisticClock@acaimingus/files/minimalisticClock@acaimingus/desklet.js
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    // Handle single digit or double digit for minutes
+    if (datem < 10) {
+      this.min1.set_text("0");
+      this.min2.set_text(datem.toString());
+    } else {
+      this.min1.set_text(datem.toString()[0]);
+      this.min2.set_text(datem.toString()[1]);
+    }
```
rcalixte, line 299: There are functions in GLib for this.

minimalisticClock@acaimingus/screenshot.png
rcalixte, line 1: This file is unable to render.

rcalixte (changes requested):

rcalixte:
@acaimingus Not all of these comments are blockers but there are a few troublesome things here. The irony is that almost all of this functionality already exists in another desklet plus more, `dual-datetime@rcalixte`. (I opted to leave font installation and choice up to the user.) Take a look at the code but to be frank, you can accomplish most of what you've done here with it, certainly all of the core functionality. The Spices page for it is [here](https://cinnamon-spices.linuxmint.com/desklets/view/64) as well.

acaimingus:
Alright, it's probably better then to use that desklet instead of making a new one. Thank you for the review.

rcalixte:
> Alright, it's probably better then to use that desklet instead of making a new one.

Can you verify that it fits your needs?

acaimingus:
> > Alright, it's probably better then to use that desklet instead of making a new one.
> 
> Can you verify that it fits your needs?

Yes, it is pretty close and it properly handles scaling as well as being more robust in general. The only thing that's missing is font weight, but I guess I could make a PR for that. So I think it's a better idea to focus my efforts on suggesting features to that desklet instead of forcefully making my own from scratch, which doesn't work as well.

rcalixte:
> The only thing that's missing is font weight

This is handled by the input if the font supports it, e.g. Ubuntu Light or Ubuntu Bold.



> So I think it's a better idea to focus my efforts on suggesting features to that desklet instead of forcefully making my own from scratch

This is my personal take but in general, I try to find ways to extend existing things rather than starting from scratch. This pseudo-credo is ultimately very relative so it should be taken on a case-by-case basis.

PR #1285, opened 2024-09-18
dual datetime@rcalixte: Feature suggestion: Add font weight

acaimingus:
I would like to suggest adding an font weight option to the desklet. I know this is already partially handled as mentioned by you, @rcalixte:

>> The only thing that's missing is font weight
>
> This is handled by the input if the font supports it, e.g. Ubuntu Light or Ubuntu Thin. Some fonts don't respond by their
> human-readable name for all variants which is a little frustrating but it wasn't something I thought merited too much 
> investment beyond the current implementation. (You can see what is supported by running `fc-list`. You'll see that many fonts are more limited than their display names might suggest.)
>
>_Originally posted by @rcalixte in https://github.com/linuxmint/cinnamon-spices-desklets/issues/1282#issuecomment-2352805536_

However I don't think this is a satisfactory solution and here are a few reasons for it:

- With the current way it is variable weight fonts (like [SUSE](https://fonts.google.com/specimen/SUSE), [EB Garamond](https://fonts.google.com/specimen/EB+Garamond) and [many, many more](https://fonts.google.com/?vfonly=true)) are not adjustable beyond their standard normal 400 weight, as these fonts usually don't even provide names for all the possible weights, reducing options
- The font-weight option works (as far as I'm aware) on all fonts (static or variable weight)
- Using font-weight enables in many cases more granular control over the font style, which makes the desklet even more customizable
- Font-weight can also be used in combination with a specified font, which means that it would only warrant further improvement.

Please let me know your thoughts on this.


rcalixte:
> * these fonts usually don't even provide names for all the possible weights, reducing options

Isn't this an upstream problem?



> * Font-weight can also be used in combination with a specified font, which means that it would only warrant further improvement.

What do you mean here?



> Please let me know your thoughts on this.

Initially very mixed. I'll take a deeper look later.

acaimingus:
> > * these fonts usually don't even provide names for all the possible weights, reducing options
> 
> Isn't this an upstream problem?

Variable fonts have been designed to reduce disk usage and clutter by not having 30 different font files for 1 font. They are intended to be modified through font-weight and similar css-styles, so this is not really a problem, but intended behaviour. 
 
> > * Font-weight can also be used in combination with a specified font, which means that it would only warrant further improvement.
> 
> What do you mean here?

 This is a bit of a repeated point, my bad. Essentially what is meant with this is that font-weight maps static fonts to the font-weight range and picks the font with the value closest to the chosen weight.

rcalixte:
> I would like to suggest adding an font weight option to the desklet.

I tested it and it works fine. It's also implemented in a way that is maintainable. One note, when working with the translation files, you can use the `cinnamon-spices-makepot` script to update the template file and any existing translation files rather than doing it manually. Using Poedit for the translations also ensures auto-formatting remains intact.

rcalixte (approved):

acaimingus:
> I tested it and it works fine. It's also implemented in a way that is maintainable.

Great to hear! 

> One note, when working with the translation files, you can use the `cinnamon-spices-makepot` script to update the template file and any existing translation files rather than doing it manually. Using Poedit for the translations also ensures auto-formatting remains intact.

Ah, thank you for the tip, I'll do it that way in the future.

PR #1287, opened 2024-09-21
InternetTimeDesklet@stefan: Added German translation

acaimingus:


acaimingus:
Accidentially added another commit to this, my bad.

rcalixte:
> Accidentially added another commit to this, my bad.

For the future, not a problem since we can squash them.

acaimingus:
Ah, okay, good to know.

PR #1293, opened 2024-09-28
notes@schorschii: Allow in desklet note taking

anaximeno:
Hello @schorschii,
With this PR I generally made two changes:
1. Reorganize settings, dividing it in two sections **General**, and **Visual**
2. Allow taking notes in the Desklet with the **textview** settings component type. This is a toggeable option, the default is still the text-based note, but you can edit notes directly in the desklet settings now, which makes sense since each Desklet will have its own settings instance.

Example Picture:

Captura de ecrã de 2024-09-28 06-52-54

PR #1296, opened 2024-09-29
notes@schorschii: update/fix translations

schorschii:
- removed the 'fuzzy' flag which prevents translations from being displayed
- fixed wrong translations introduced with #1293

### DIFF
--- a/notes@schorschii/files/notes@schorschii/po/ca.po
+++ b/notes@schorschii/files/notes@schorschii/po/ca.po
@@ -68,14 +68,12 @@ msgid "Note Taking Method"
 msgstr ""
 
 #. settings-schema.json->note-taking-method->options
-#, fuzzy
 msgid "Text File"
-msgstr "Color del text"
+msgstr "Fitxer de text"
 
 #. settings-schema.json->note-taking-method->options
-#, fuzzy
 msgid "Desklet Settings"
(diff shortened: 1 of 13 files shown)


rcalixte:
> * removed the 'fuzzy' flag which prevents translations from being displayed

These strings are inserted to represent that the translations should be double-checked. They are not prevented from being displayed. This notation allows tools with proper support to identify these fields for translators as well.

Going forward, can you make use of the repository scripts for translations or use Poedit and avoid editing them manually? Even Vim has plugins for translation file support if you would rather go that route.

rcalixte (approved):

schorschii:
> These strings are inserted to represent that the translations should be double-checked. They are not prevented from being displayed.

My experience was different. I wondered why [these](https://github.com/linuxmint/cinnamon-spices-desklets/pull/1296/files#diff-0d94ba9734759268e959f4f49517d0bee86a53ab37d1d9254274d316e6c08166L68) German translations were not displayed. As soon as I removed the fuzzy flag and executed `cinnamon-spices-makepot notes@schorschii`, they were displayed correctly (yes, I'm aware of the repository scripts).

rcalixte:
> My experience was different.

There have been other translation issues with your setup before right? Have those been resolved?

schorschii:
I didn't notice any translation issues before.

PR #1343, opened 2024-11-28
temperature@swalladge: consistent spaces in translations like in the .pot

KopfdesDaemons:


rcalixte:
This isn't proper punctuation in every language is it?

KopfdesDaemons:
Okay, but then the Spanish translator made a typo, because there it is not consistent within the Spanish translation. Even in English there is a space (.pot).

PR #1381, opened 2025-01-07
[googleCalendar@javahelps.com] Improved compatibility with Cinnamon 6.4 - Bugfixes

claudiux:
@slgobinath 
These changes avoid errors on Cinnamon 6.4, by avoiding the use of `Mainloop.source_remove()` and adding the _Open Google Calendar_ menu item when it is sure that _this._menu_ exists.


slgobinath:
Thanks @claudiux 
Looks good to me

claudiux:
@slgobinath 
Mmh. It seems to me that there is a problem changing the _delay_ value. I'll check it soon.

claudiux:
@slgobinath 
All is fine, now.

PR #1394, opened 2025-02-06
system-monitor-graph@rcassani: Add support for amdgpu, and bump version

Lazar-Kovacevic:
This commit adds support for showing GPU and VRAM usage on GPUs which make use of the amdgpu driver.

It should be noted that GPU usage can be somewhat inaccurate because of the low sampling frequency.

### DIFF
--- a/system-monitor-graph@rcassani/files/system-monitor-graph@rcassani/desklet.js
+++ b/system-monitor-graph@rcassani/files/system-monitor-graph@rcassani/desklet.js
@@ -251,9 +251,35 @@ SystemMonitorGraph.prototype = {
                             break;
                     }
                     break;
-
+                case "amdgpu":
+                    switch (this.gpu_variable) {
+                        case "usage":
+                            this.get_amdgpu_gpu_use();
+                            value = this.gpu_use / 100;
+                            text1 = _("GPU Usage");
+                            text2 = Math.round(this.gpu_use).toString() + "%";
+                            break;
+                        case "memory":
+                            this.get_amdgpu_gpu_mem();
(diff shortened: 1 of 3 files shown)


rcalixte:
cc @rcassani

rcassani:
Thanks @Lazar-Kovacevic for adding the support for AMD, that is great.

Currently in the code, the `switch` structures inside the GPU case are nested below, then the code indicated with the `(*)` is duplicated:

```
SWITCH GpuManufacturer
   CASE nvidia
   SWITCH GpuVariable
       CASE usage
           GET nvidia_usage 
           PREPARE usage label (*)
       CASE memory
           GET nvidia_mem
           PREPARE mem label (*)

   CASE amd
   SWITCH GpuVariable
       CASE usage
           GET amd_usage
           PREPARE usage label (*)
       CASE memory
           GET amd_mem
           PREPARE mem label (*)      
```

@rcalixte and @Lazar-Kovacevic
A better approach would be to change the order of the `switch` structures as indicated below.
With this change, expanding to more GPU manufactures will be straight forward, and code will not be duplicated.

```
SWITCH GpuVariable
    CASE usage
        SWITCH GpuManufacturer
            CASE nvidia
                GET nvidia_usage 
            CASE amd
               GET amd_usage
            PREPARE usage label

    CASE memory
        SWITCH GpuManufacturer
            CASE nvidia
                GET nvidia_mem 
            CASE amd
               GET amd_mem
            PREPARE mem label
```

Lazar-Kovacevic:
Done @rcassani

rcassani:
@Lazar-Kovacevic, thanks for the quick update
@rcalixte, it looks good to me

rcalixte (approved):

PR #1398, opened 2025-02-09
yarr 1.4.1

jtoberling:
Pls - as new branch.

Also added some slight fixups.


rcalixte (approved):

rcalixte:
@jtoberling Nice job, just a note that there is a script in the top level of the repository named `cinnamon-spices-makepot` that you can run to update the translation files. Think of it as a final step, as the line numbers from code and strings from the settings are pulled in at run-time so you should do it when you're set with the final code.

Issue #1396, opened 2025-02-09
[system-monitor-graph@rcassani] Desklet crashes


jross365:
### Desklet version/Build date

1.7

### Cinnamon version

6.4.6

### Distribution

Mint 22.1

### Graphics hardware *and* driver used

Nvidia RTX 3080 and 550.120

### Desklet name and maintainer

system-monitor-graph@rcassani @rcassani

### What happened?

After upgrading to the latest version of System Monitor Graph, Nvidia GPU Usage/VMEM graph desklets crash. After rebooting system, Cinnamon reports an applet crashed.

After removing/re-adding the GPU desklets, they both sit on a non-functional "CPU" graph.

jarama:
Yes, same here (also using NVIDIA). After removing the GPU panels I had trouble adding other panels as well. It seems they refuse to appear. After the update only the CPU panel remained visible from the six panels that I had (CPU, GPU usage, GPU mem, RAM, HDD 1, HDD 2).

jross365:
**Update:** These corresponding errors are also in my "LookingGlass" log (_/home/username/.xession-errors_ and _/home/username/.xsession-errors.old_)

> js-Message: 22:07:03.379: JS LOG: [LookingGlass/error] [Gio.IOErrorEnum: Error opening file /sys/class/drm/card0/device/gpu_busy_percent: No such file or directory]: Failed to evaluate 'main' function on desklet: system-monitor-graph@rcassani/3
> logError@/usr/share/cinnamon/js/ui/extension.js:143:17
> Gjs-Message: 22:07:03.386: JS LOG: [LookingGlass/error] [Gio.IOErrorEnum: Error opening file /sys/class/drm/card0/device/gpu_busy_percent: No such file or directory]: Failed to evaluate 'main' function on desklet: system-monitor-graph@rcassani/3
> logError@/usr/share/cinnamon/js/ui/extension.js:143:17
> Gjs-Message: 22:07:03.392: JS LOG: [LookingGlass/error] [Gio.IOErrorEnum: Error opening file /sys/class/drm/card0/device/gpu_busy_percent: No such file or directory]: Failed to evaluate 'main' function on desklet: system-monitor-graph@rcassani/3
> logError@/usr/share/cinnamon/js/ui/extension.js:143:17
[...]

PR #1399, opened 2025-02-09
Bugfix: `switch` structures in JavaScripts are fall-through

rcassani:
Fixes the bug reported in #1396 

The bug was introduced in #1394 as `switch` structures in JavaScript are fall-through. 
https://www.freecodecamp.org/news/fall-through-in-javascript-switch-statements/

So, with `NVIDIA` selected, both cases `nvidia` and `amd` were executed.
The fix just ends each GPU manufacturer case with `break;`

Also
- Updated version number
- Indentation change for clearer code

@Lazar-Kovacevic, the AMD support should be working properly

### DIFF
--- a/system-monitor-graph@rcassani/files/system-monitor-graph@rcassani/desklet.js
+++ b/system-monitor-graph@rcassani/files/system-monitor-graph@rcassani/desklet.js
@@ -228,8 +228,10 @@ SystemMonitorGraph.prototype = {
                       switch (this.gpu_manufacturer) {
                           case "nvidia":
                               this.get_nvidia_gpu_use();
+                              break;
                           case "amdgpu":
                               this.get_amdgpu_gpu_use();
+                              break;
                       }
                       value = this.gpu_use / 100;
                       text1 = _("GPU Usage");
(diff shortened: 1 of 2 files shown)


rcalixte (approved):

rcalixte:
Sorry, I should have double-checked while reviewing. I can never fully remember which programming languages fall through and which ones don't. Branches of `if-else-if` aren't necessarily more elegant but they're how I tend to avoid this issue when I'm moving between a lot of programming languages in a short period of time. I still should have double-checked before merging since I wasn't really sure. Apologies.

jross365:
Thanks for the quick fix!

rcassani:
@rcalixte, same here, I just eye-balled rather than test the PR. 
It is just an extra case, it should be OK 🤣

This is my only project in JS, thus I forget all the time this small changes.

Lazar-Kovacevic:
I am so sorry for the mess everyone, i should have paid more attention.

I don't have any nv gpus, so as it too often happens, nothing seemed wrong on my setup.

I really hope nobody lost any important work due to this and again, i'm really sorry.

PR #1406, opened 2025-02-13
lastx@swud: Initial submission

swudhead:


lastx@swud/info.json
```diff
@@ -0,0 +1,3 @@
+{
+    "author": "swud"
```
rcalixte: Just a note that this should be your GitHub username.

PR #1417, opened 2025-03-05
[moonlight-clock@torchipeppo] Initial release

torchipeppo:
Proposing a desklet I made to combine visibly displaying the current date & time and keeping track of imminent important events, with more features I thought would be useful in general and/or fun to have.


claudiux:
This error message appears in `~/.xsession-errors`:
```
Gjs-Message: 11:46:27.151: JS LOG: [LookingGlass/error] [undefined]: File not found: /home/claudiux/.local/share/cinnamon/desklets/moonlight-clock@torchipeppo/5.7/desklet.js
Gjs-Message: 11:46:27.151: JS LOG: [LookingGlass/trace] 
<----------------
logError@/usr/share/cinnamon/js/ui/extension.js:143:17
ensureFileExists@/usr/share/cinnamon/js/ui/extension.js:179:15
finishLoad@/usr/share/cinnamon/js/ui/extension.js:221:29
_init/</<@/usr/share/cinnamon/js/ui/extension.js:269:24
---------------->
```

rcalixte:
I get:

```
Gjs-Message: 06:25:36.269: JS LOG: [LookingGlass/error] 
[devtest-moonlight-clock@torchipeppo]: imports.ui.deskletManager.deskletMeta[UUID] is undefined
[devtest-moonlight-clock@torchipeppo]: Error importing desklet.js from devtest-moonlight-clock@torchipeppo
Gjs-Message: 06:25:36.269: JS LOG: [LookingGlass/trace] 
```

Use the `test-spice` script in the root of the repository to test this. It looks like we could update the validation script with a few more conditions too.

rcalixte:
I think you should just have the folder be either 5.6 or 5.8 (since those are actually Cinnamon versions) and then remove the files from the top-level. The layout needs to be one or the other. When there is a versioned folder, just default to that unless there are multiple folders.

rcalixte:
@torchipeppo I would also suggest moving the assets. Either have them inside the desklet directory if you want them included or put them at the top level of the folder before `files` if you do not.

torchipeppo:
> @rcalixte first message (a different error message)
Okay, that I know why: I am hardcoding the UUID in order to get the desklet path, as I've seen do in multiple other spices, and make the "old" import system work.

> @rcalixte second message (versions)
I'll have to check again how I came to choose the 5.7 number then. The reason is the adoption of TextDecoder for reading files, btw.

> @rcalixte third message (assets)
Isn't that already true? The screenshots are only for the README, so they are outside `files`. The SVG is part of the desklet, and it's in the desklet directory.

torchipeppo:
Alright, I'll need a little time to sort this all out.
Double version will probably still be a thing because I'd like this to also work on my old Mint 20 machine too (until I finally find the time to backup and upgrade, anyway).
But I think I noticed that, even there, it switched to the new import system once I added the if-branch, so I'll have to confirm.
Worst case, if I decide it's too hard to maintain both, I'll just make this Cinnamon 6+ only and keep a separate, personal "legacy" version just for the other machine.

torchipeppo:
Hi, I have a new release ready for review. validate-spice says no error, and test-spice looks good on both of my machines.

> @rcalixte imports.ui.deskletManager.deskletMeta[UUID] is undefined

Eliminated the old import system and all harmful hardcoded UUIDs. Turns out the new one is okay in Cinnamon 5 too. The only ones I couldn't handle are those used to query gettext for the translations, which appear to be harmless anyway (tested with test-spice), so worst case, translations silently won't work in devtest mode. Is this acceptable?

> @rcalixte version folders

Found the actual version where TextDecoder was introduced and renamed the version folder accordingly. Also moved old-version files away from the top-level, creating another version folder which I named after the version where I believe the new `require`-based import system was introduced.

> @claudiux File not found: /home/claudiux/.local/share/cinnamon/desklets/moonlight-clock@torchipeppo/5.7/desklet.js

Could not reproduce your error yet, and I don't currently have a way to test other environments, such as fresh installs or other Cinnamon versions. I can only hope that taking care of the other compatibility issues indirectly fixed yours too. Could you tell me if the error persists on your end?

rcalixte:
> Eliminated the old import system and all harmful hardcoded UUIDs. Turns out the new one is okay in Cinnamon 5 too. The only ones I couldn't handle are those used to query gettext for the translations, which appear to be harmless anyway (tested with test-spice), so worst case, translations silently won't work in devtest mode. Is this acceptable?

Oh yeah, I don't think that was a test path for the scripts, just core functionality.

rcalixte (approved):

rcalixte:
Very nice, just a note that the additional comments in the translation template won't persist past additional script executions. (In general, hand-modifying them isn't a good practice.) It should be fine in the short-term though.

torchipeppo:
That's what I thought too. The only additional comment is a pointer to the README included in the `po` folder, right?

PR #1435, opened 2025-03-25
[moonlight-clock@torchipeppo] Add a README file for translation context

torchipeppo:
This should have been released since the beginning, but it appears I missed it until a few days ago. Anyway, here it is.


rcalixte:
Maybe name it TRANSLATORS.md in the directory above or append a "For translators" section in the README?

torchipeppo:
Will do, thanks

PR #1457, opened 2025-04-20
Fix issue 1456

sohail-bakhtiari:
@mohammad-sn 

It is a more complete fix to include all unicode characters rendering.
I have also moved the replace command down to include poet name (just in case).


rcalixte:
@sohail-bakhtiari I think you need to either rebase this branch or create a new branch that has the updates from your earlier pull request. This change appears to be out of sync.

rcalixte (approved):

sohail-bakhtiari:
> @sohail-bakhtiari I think you need to either rebase this branch or create a new branch that has the updates from your earlier pull request. This change appears to be out of sync.

Sorry, I thought making a `pull` in that branch would be enough.
Thank you for fixing it.

PR #1511, opened 2025-06-27
Add Network Monitoring Support to System Monitor Graph

FSSCoding:
# Add Network Monitoring Support to System Monitor Graph Desklet

## 📋 Summary

This pull request adds comprehensive **network monitoring** functionality to the existing system monitor graph desklet. Users can now monitor real-time upload and download speeds with a dual-line graph display, maintaining full compatibility with all existing features.

### **Network Monitoring Type**
- **Dual-line graph**: Separate visualisation for download (green) and upload (orange/brown) speeds
- **Real-time data**: Reads from `/proc/net/dev` for accurate network statistics
- **Auto-scaling**: Graph dynamically adjusts scale based on peak network activity
- **Interface selection**: Monitor all network interfaces or specify individual ones (eth0, wlan0, etc.)
- **Smart formatting**: Automatic unit conversion (B/s → KiB/s → MiB/s → GiB/s)

### **Configuration Options**
- **Network type**: Added "Network" option to monitoring type dropdown
- **Interface filter**: Optional text field to specify which network interface to monitor
- **Unit prefixes**: Choice between Binary (1024-based) and Decimal (1000-based) units
- **Dual colours**: Separate colour pickers for download and upload lines
- **All existing settings**: Full compatibility retained

#### 📌 Monitoring Type Option
```json
"type": {
  "type": "combobox",
  "default": "cpu",
  "options" : {
    "CPU" : "cpu",
    "RAM" : "ram",
    "Swap" : "swap",
    "HDD" : "hdd",
    "GPU" : "gpu",
    "Network" : "network"
  },
  "description": "Select what type of system resource to monitor"
}
```

#### 📌 Network-Specific Settings
```json
"data-prefix-network": {
  "type": "combobox",
  "default": 0,
  "options": {
    "Binary prefix or IEC (1 KiB = 1024 bytes)" : 0,
    "Decimal prefix (1 KB = 1000 bytes)" : 1
  },
  "description": "Prefix for data units",
  "tooltip": "Unit prefix for network speed.",
  "dependency": "type=network"
},
"network-interface": {
  "type": "entry",
  "default": "",
  "description": "Network interface to monitor",
  "tooltip": "Leave empty to monitor all interfaces. Example: eth0, wlan0",
  "dependency": "type=network"
},
"line-color-network-down": {
  "type": "colorchooser",
  "default": "rgba(100,180,120,1.0)",
  "description": "Line colour Network Download",
  "tooltip": "RGBA supported; alpha affects fill/line transparency",
[...]

PR #1531, opened 2025-07-31
 system-monitor-graph@rcassani: network monitoring adjustments

schorschii:
- I think it makes more sense to display network speed as bits/sec rather than bytes/sec, so I added an option for that and made it the default.
- The up/download speed sometimes overlaps, depending on the amount of chars of the current network speed. That's why, I moved this info into the smaller, right-aligned `text3` label.
- Furthermore, I fixed some German translations.

What do you think @rcassani ?

### DIFF
--- a/system-monitor-graph@rcassani/files/system-monitor-graph@rcassani/settings-schema.json
+++ b/system-monitor-graph@rcassani/files/system-monitor-graph@rcassani/settings-schema.json
@@ -61,7 +61,7 @@
         "options": {
             "Bytes - Binary prefix or IEC (1 KiB = 1024 bytes)" : 0,
             "Bytes - Decimal prefix (1 KB = 1000 bytes)" : 1,
-            "Bits (1 Kb = 1000 bit)" : 2
+            "Bits - Decimal prefix (1 Kb = 1000 bit)" : 2
         },
         "description": "Prefix for data units",
         "tooltip": "Unit prefix for network speed.",
--- a/system-monitor-graph@rcassani/files/system-monitor-graph@rcassani/settings-schema.json
+++ b/system-monitor-graph@rcassani/files/system-monitor-graph@rcassani/settings-schema.json
@@ -61,7 +61,7 @@
         "options": {
             "Bytes - Binary prefix or IEC (1 KiB = 1024 bytes)" : 0,
             "Bytes - Decimal prefix (1 KB = 1000 bytes)" : 1,
-            "Bits - Decimal prefix (1 Kb = 1000 bit)" : 2
+            "Bits - Decimal prefix (1 Kb = 1000 bits)" : 2
         },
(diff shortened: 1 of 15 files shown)


rcassani:
> I think it makes more sense to display network speed as bits/sec rather than bytes/sec, so I added an option for that and made it the default.

That's a nice feature.
Just do not make the default. Similar as it is done within the GNOME System monitor, `Bytes/s` are default, but user can turn on the option to have `bits/s`. Also, it should play nicely with the prefixes (binary and decimal) to have kilobits and kibibits.

> The up/download speed sometimes overlaps, depending on the amount of chars of the current network speed. That's why, I moved this info into the smaller, right-aligned text3 label.

Is this overlapping related to the change of units? 
In the PR that introduced the network option, the size of the up/down text was discussed:
https://github.com/linuxmint/cinnamon-spices-desklets/pull/1511#pullrequestreview-2969003192

> Furthermore, I fixed some German translations.
Really appreciated 🙏

system-monitor-graph@rcassani/files/system-monitor-graph@rcassani/settings-schema.json
```diff
@@ -59,8 +59,9 @@
         "type": "combobox",
         "default": 0,
         "options": {
-            "Binary prefix or IEC (1 KiB = 1024 bytes)" : 0,
-            "Decimal prefix (1 KB = 1000 bytes)" : 1
+            "Bytes - Binary prefix or IEC (1 KiB = 1024 bytes)" : 0,
+            "Bytes - Decimal prefix (1 KB = 1000 bytes)" : 1,
+            "Bits (1 Kb = 1000 bit)" : 2
```
rcassani: For sake of coherency, it should be:
`"Bits - Decimal prefix (1 Kb = 1000 bits)"`

rcassani (changes requested):
It looks good! 
One minor change is requested.

---
I'm curious about [7a31cf0](https://github.com/linuxmint/cinnamon-spices-desklets/pull/1531/commits/7a31cf02afc93f0f822b314ccf956bc21975f51b), is there a list of fields for the `metadata.json`?

rcassani:
After some tests, I agree the small text looks better.
FYI @FSSCoding

> No, I noticed the overlapping issue before my change. Sorry for mixing this up in one PR.

There is no problem, several improvements can the in the same PR.

rcassani:
@schorschii, thanks for the info about `metadata.json`

About the change in [9f5b323](https://github.com/linuxmint/cinnamon-spices-desklets/pull/1531/commits/9f5b323d00371082c191246a15cdf89333f94e6a), it should be `bits` instead of `bit`

rcassani (approved):
@schorschii, thanks!
@rcalixte, this is ready to be merged!

PR #1581, opened 2025-09-24
nvidia-gpu-monitor@hierakonpolis: a desklet plotting GPU usage info over time, memory, compute and temperature

Hierakonpolis:
A real-time GPU monitoring desklet that displays compute usage, memory usage, and optional temperature with historical graphs. Features dual Y-axis display with percentage scale on the left and temperature scale (30-90°C) on the right. Uses nvidia-smi for data collection and update intervals can be configured.


Hierakonpolis:
@claudiux As far as I can tell, yes

PR #1598, opened 2025-10-17
calendar@deeppradhan: Add primary display setting

### LINKED ISSUE #1447: Desklet Feature Request - Make clock swappable with current day

calendar@deeppradhan @kanchudeep

### What would you like to see?

I'd like to see an option, where I can set the big number for clock instead, and push the day down in the place of the small texted clock, so basically make a toggle button, which allows the user, to decide whether they want the current day to be the Big bold text, or the Clock.

here image for better understanding

Image

KopfdesDaemons:
This closes a feature request. I've added a setting to adjust the primary display. The primary display will be displayed with larger font and will always be in second position.

closes #1447


fredcw:
@kanchudeep

PR #1632, opened 2025-11-05
TimeAndDate@nightflame: V2.0 Overhaul

NotSirius-A:
Hello, I've noticed this desklet also has not been receiving updates for a while. This desklet has 17 stars now and it's useful, so I think it's safe to assume that some people are using it now or new users will be trying it out, so these changes are needed. 

Customization of this desklet (manually editing `metadata.json` and `stylesheet.css`) was very limited and also difficult for people how didn't know CSS. I added lots of customization options to bring this one to modern standards/expectations. Unfortunately, these changes will destroy previous setups users might have, but I don't think I can do anything about that.

screenshot


fredcw:
Super duper fabulous. LGTM

@rcalixte @claudiux

PR #1637, opened 2025-11-07
test-spice: Added option to replace uuid to devtest-uuid in copied desklet.js

NotSirius-A:
Hello, I updated the `test-spice` script. I added an option to automatically prepend the `devtest-` prefix inside of the copied `desklet.js` file. 

I made this change because it was annoying me that I had to remove/add the `devtest-` prefix to UUID inside `desklet.js` every time I made a new commit. This is relevant when desklets use some extra files and need to refer to their own directory. I suspect I'm not the only one doing this. 

I'm not sure if I'm allowed to make such changes. Please let me know if not.

### DIFF
--- a/test-spice
+++ b/test-spice
@@ -48,11 +48,10 @@ def copy_xlet(uuid, change_uuid=False):
 
         if change_uuid:
             desklet_js_file = f'{dev_dest}/desklet.js'
-            with open(desklet_js_file, 'r') as file:
+            with open(desklet_js_file, 'r+') as file:
                 js_code = file.read()
-                
-            js_code = js_code.replace(uuid, f"{DEV_PREFIX}{uuid}")
-            with open(desklet_js_file, 'w') as file:
+                file.seek(0)
+                js_code = js_code.replace(uuid, f"{DEV_PREFIX}{uuid}")
                 file.write(js_code)
 
         reload_xlet(f'{DEV_PREFIX}{uuid}')


KopfdesDaemons:
I am also aware of the problem. So far I have also always had to rename the UUID in the desklet.js.

test-spice
```diff
@@ -45,9 +45,19 @@ def copy_xlet(uuid):
...
             json.dump(data, metadata, indent=4)
+
+        if change_uuid:
+            desklet_js_file = f'{dev_dest}/desklet.js'
+            with open(desklet_js_file, 'r') as file:
+                js_code = file.read()
+                
+            js_code = js_code.replace(uuid, f"{DEV_PREFIX}{uuid}")
+            with open(desklet_js_file, 'w') as file:
+                file.write(js_code)
```
rcalixte, line 55:
```suggestion
        if change_uuid:
            desklet_js_file = f'{dev_dest}/desklet.js'
            with open(desklet_js_file, 'r+') as file:
                js_code = file.read()
                file.seek(0)
                js_code = js_code.replace(uuid, f"{DEV_PREFIX}{uuid}")
                file.write(js_code)
```

NotSirius-A:
> This was intentional at the time because I was wary of modifying developer files. I believe it's possible to dynamically use the directory path rather than having it hard-coded on the UUID which would solve these problems but I need to double-check that to be sure. If so, it would be more about better practices. If not, this change makes more sense.

Ok, I understand, makes sense. If it is possible to set the directory dynamically for devtest desklets then please let us know. I don't think I've seen it anywhere. Of course this PR would be unnecessary then.

So far, I've been using these two lines:
```js
const UUID = "todo@NotSirius-A";
const DESKLET_ROOT = imports.ui.deskletManager.deskletMeta[UUID].path;
```

PR #1676, opened 2025-11-22
Cinamp: New Desklet

mlandsjr:
Added my Cinamp Desklet.


fredcw:
desklet doesn't load for some reason, I don't know why, nothing in ~/.xsession-errors. Also, directory structure is wrong: https://github.com/linuxmint/cinnamon-spices-desklets?tab=readme-ov-file#file-structure

mlandsjr:
fixed icon size

claudiux:
Volume bar and Progress bar do not display nothing.
Icon is not squared.
image

PR #1709, opened 2025-12-21
Cinnaclock: New layered clock desklet

Dingo-linux:
Hello, I would like to submit a new desklet called **Cinnaclock**.

It is a modern and highly customizable clock that displays time, day, and date on separate layers. 

**Key features:**
- Independent visibility and styling for Time, Day, and Date.
- Custom fonts, colors, and sizes for each element.
- Adjustable vertical offsets to create unique overlapping typography.
- Customizable text shadows and uppercase toggles.
- Fully translatable (POT and several PO files included).

I have included the required `metadata.json`, `icon.png`, and `screenshot.png` files.


fredcw:
Your applet shouldn't be nested inside the "desklets" folder in this PR. Apart from that, the file structure is good. The validation tool doesn't like the "icon" field in metadata.json. "author" field is not needed in metadata.json as it's already in info.json. You should probably add `"max-instances": -1,` to metadata.json.

PR #1719, opened 2026-01-04
serverstatus@maksmokriev: updates

sulonetskyy:
Updates:
- fixed max-instances
- removed configurable icons 
- changed image-icons to symblolic-icons
- removed status-postfix (-OK, -ERROR) as status-icon and status-color looks like enough
- added configurable font size
- added configurable fixed widget size
- added default server as an example for start
- added predefined options for Type-Of-Check instead of manual input
- updated translation files


claudiux:
Ping @maksmokriev

PR #1829, opened 2026-04-27
Prevents color overriding when the desklet is reloaded

mortim:
I noticed that the ``this.color = 'rgb(255,255,255)';`` basically overrides the color from the desklet configuration file. Thus, when I reload the desklet, it sets the current color to ``rgb(255,255,255)`` in the configuration.


fredcw:
Ping @ccongo

btw @ccongo, the author field in `info.json` should be your github name.

mortim:
> Ping @ccongo
> 
> btw @ccongo, the author field in `info.json` should be your github name.

@fredcw I think the real author is mkessas. His Github account doesn't exist anymore but you can see some of his replies (he's named "Mourad" btw) from the cinnamon-spices website.

https://cinnamon-spices.linuxmint.com/desklets/view/77.

Issue #820, opened 2022-12-19
kdecdesklet@joejoetv didn't work


christian179717:
```
 * Version 0.1 (2022-06-03 13:18:42)
 * Cinnamon 5.4.12
 * Mint 21
 * Intel Corporation HD Graphics 5500
 * Snippet or attachment of ~/.cinnamon/glass.log not present
 * Snippet or attachment of ~/.xsession-errors nothing related to that

 ```

**Notify author of desklet**
@JoeJoeTV 

**Issue**
Nothing meaningful shown in the desklet
![grafik](https://user-images.githubusercontent.com/89655752/208458334-2d65d543-c731-4fa4-9b55-90abf69e76e3.png)

**Steps to reproduce**
- connect your device
- choose the device in the available devices field (available via right click)

**Expected behaviour**
information about the device should be shown

**Other information**
KDE-Connect-version: kdeconnect-cli 21.12.3 (in terminal kdeconnect-cli -v)

JoeJoeTV:
Thanks for reporting!

This issue is because the KDEConnect version included in Linux Mint 21 repos changed some of the interfaces.
I plan on rewriting the desklet when I have time and I'll try to make it compatible with the old and new interface.

christian179717:
No Problem

thanks for that.

christian179717:
In principle, I would have the cunning and time to help you. Unfortunately, I have no idea about Javascript. If you want, we could see together if we can get it working again.

JoeJoeTV:
I've already started planning how I'm going to rewrite the code to be more modular and support multiple versions. I may push a fix sooner to fix the current version, if it's just replacing dbus paths, but maybe I'll just push the rewrite once I'm done. I will start soon I think.

JoeJoeTV:
Just to give an update tp this:
I have already rewritten the KDEConnect Applet and have started a bit with the Desklet rewrite.
The rewrite of the Applet will probably help me a lot in rewriting the Desklet.
I haven't made much progress there yet, since I first wanted to rewrite the Applet and was busy in between.

christian179717:
I've changed too KDE as desktop environment so no need for me anymore.

fredcw:
@JoeJoeTV Do you still plan to update this desklet. Perhaps it should be removed if not?

JoeJoeTV:
> @JoeJoeTV Do you still plan to update this desklet. Perhaps it should be removed if not?

[...]

PR #1840, opened 2026-05-04
[kdecdesklet@joejoetv] Remove desklet as it's currently non-functioning

fredcw:
With consent of author: https://github.com/linuxmint/cinnamon-spices-desklets/issues/820#issuecomment-4371624722


rcalixte:
I'll takeover the maintenance since I've been using it and patching it locally.

PR #1849, opened 2026-06-04
clock@schorschii: adjust README for pro version

schorschii:


mtwebster:
Hi,

We maintain the Spice repo and 'store' in order to provide some oversight and protection for users so they feel comfortable installing Spices.

Providing a 'premium version' (which encourages downloading files outside of this ecosystem) breaks this trust, and potentially exposes users to things that are out of our control.

We'll be updating our README here to clarify this, and avoid confusion in the future.

Thank you.

PR #1857, opened 2026-06-18
yfquotes@thegli - weekly and monthly change

v-radev:
This PR adds options to enable extra columns for weekly change and monthly change of each stock. The display value can be configured to be 1 of 3 options: percent change, absolute change (difference) or just display the market price from a week/month ago.

The data required for this is fetched with a single request if the tickers are up to 20, if they are more than 20, the requests are chunked to 20 tickers each, which is the limit forced by YF. 

If one of the weekly or monthly options are enabled, the data is merged to the existing quotes data argument and then used when the labels are getting created. I think merging the data introduces the least amount of changes to existing methods and doesn't require any new arguments being passed around.

image
image
image

### DIFF
--- a/yfquotes@thegli/CHANGELOG.md
+++ b/yfquotes@thegli/CHANGELOG.md
@@ -1,3 +1,10 @@
+### 0.18.4 - June 18, 2026
+
+Features:
+
+- option to display columns for weekly and monthly change for each stock
+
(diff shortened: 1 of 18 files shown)


v-radev:
@thegli please review whenever possible. Thanks

PR #1858, opened 2026-06-19
fix: infinite loop spikes CPU at midnight

### LINKED ISSUE #1547: Desklet freezes entire pc daily at 12am
### Desklet version/Build date

last-edited: 1754312180 

### Cinnamon version

6.4.8

### Distribution

Mint 22.1

### Graphics hardware *and* driver used

NVIDIA GeForce GTX 1660 SUPER and nvidia-driver-550 

prayertimes@linworx.org @mkessas

### What happened?

This desklet causes my pc to freeze every night at 12 am exactly. After my PC freezes, it basically now responds to every keyboard press or mouse movement after 5 seconds and its super laggy. It remains that way until I restart cinnamon. I knew it was this desklet that did it because when i removed it the issue was gone. 

### Other information

_No response_

abdelhalim97:
closes [#1547](https://github.com/linuxmint/cinnamon-spices-desklets/issues/1547)
### Root cause
`setupUI` calls refresh.
`refresh` set a timer to call `setupUI` every 1 second.
This resulted infinite loop.

At midnight this gets worst, because date changes, which will trigger `calculatePrayerTimes` and `calculatePrayerTimes` also calls `setupUI`...
Multiple infinite loops explains cpu spike at midnight.

### Fix
Replaced refresh with a function `startUpdateTimer` that run once per second.
Added updateCountdown checks if day changed and recalculate prayer times at midnight.
UI will only be built when needed instead of every second

Same features without infinite loops

### DIFF
--- a/prayertimes@linworx.org/files/prayertimes@linworx.org/5.8/desklet.js
+++ b/prayertimes@linworx.org/files/prayertimes@linworx.org/5.8/desklet.js
@@ -41,7 +41,7 @@ class MuslimPrayerTimesDesklet extends Desklet.Desklet {
 		this.settings.bind("desklet_size", "desklet_size", this.setupUI);
 		this.settings.bind("color", "color", this.setupUI);
 		this.settings.bind("show_city", "show_city", this.setupUI);
-		this.settings.bind("show_countdown", "show_countdown", this.setupUI);
+		this.settings.bind("show_countdown", "show_countdown", this.updateCountdown);
 		this.settings.bind("show_table", "show_table", this.setupUI);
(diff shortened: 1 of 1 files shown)


M-Hassand:
Great work!

In case the original author does not wish to change the original design, it is worth noting that removing the call to `setupUI` at the end of `calculatePrayerTimes` solves the problem as well (although of course, not as elegantly as the one above).

It is not needed as before the `calculatePrayerTimes` call, a call to `setupUI` was already scheduled. This way, there's only one loop that goes back and forth between `setupUI` and `refresh` rather than multiple loops spawning at midnight.

claudiux (approved):

PR #1860, opened 2026-06-27
for clock@schorschii - adding some more clock styles for analog chronometer

OnlineLearningTutorials:
some more clock styles and default clock size at startup be 250px.
This is my personal preferences as 150px clock looks small.


rcalixte:
cc @schorschii

schorschii:
Dear @OnlineLearningTutorials I appreciate your work, but I need to be honest that I'm not a fan of some of the new clock faces. E.g. the pink, blue and green clock faces are way too bright. Please don't use such glaring colors. I'd like to have a few minimalist and decent clock faces in this desklet; less is more.

And please don't change the default size. It aligns perfectly with other desklets like Battery, Calendar, Disk Space etc.

OnlineLearningTutorials:
@schorschii Thanks for your opinion, reverting to default clock size, that's my personal preferences. I am thinking of giving a background color option, which can be set to transparent or opaque, while the front clock face is transparent. Also trying to display the time zone, when multiple instance of clock placed in desktop with different time zone, it is very nice to show below which time zone each clock follows.

PR #1883, opened 2026-08-08
hostcheck: fix sizing and height calculation for status label

Heckie75:
Fix [hostcheck@schorschii](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) desklet sizing by removing incorrect global.ui_scale scaling from label dimension calculations, using the label’s preferred rendered height, and making the canvas height match the actual label height; also add box-sizing: border-box to the status box stylesheet so the layout behaves consistently.

### DIFF
--- a/hostcheck@schorschii/files/hostcheck@schorschii/desklet.js
+++ b/hostcheck@schorschii/files/hostcheck@schorschii/desklet.js
@@ -225,17 +225,21 @@ MyDesklet.prototype = {
 	showStatus: function() {
 		// refresh desklet content
 		// calc new sizes based on scale factor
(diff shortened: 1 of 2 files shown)


rcalixte:
cc @schorschii

schorschii:
I am not convinced by the change. The desklet now has a too big padding at the bottom (not equal with top padding):
Bildschirmfoto vom 2026-08-09 17-39-38

What exactly are you trying to fix?

Heckie75:
@schorschii Actually the change fixes the broken design in case that scaling is configured. In my case I have a 4K screen and set the scaling to 125%. Without this fix it looks like this:
grafik

The red/green panel is too big. Also the font is far to big.

PR #1896, opened 2026-08-25
yfquotes@thegli: refactoring

sulonetskyy:
- extract logging in a separate utility class
- use extracted logger instead of YahooFinanceQuoteUtils
- some minor fixes


sulonetskyy:
@thegli could you please review )

